PR #9, opened 2026-06-14
Add ruby.wasm in-browser playground build and re-evaluate ADR-29

zonuexe:
## Why

[ADR-29](docs/adr/29-browser-playground.md) shelved the fully in-browser (Option A) playground behind a three-condition gate (WD6) in May 2026 — the binding blocker being no Ruby 4.0 wasm build and no `prism`/`rbs` wasm packages. The ecosystem has since moved: `ruby/ruby.wasm` ships Ruby 4.0 builds, and `rbwasm` links a bundle's C extensions from source. The blocker is now a **build-and-verify** task, not a missing-upstream one.

## What this does

**1. Re-evaluates ADR-29** ([diff](docs/adr/29-browser-playground.md))
- Dated amendment + **WD6 rewritten with per-condition status**: ① Ruby 4.0 met in practice; ② `prism` met / `rbs` is the gating spike; ③ POSIX surface mapped (one `flock` site + the already-degrading fork pool + `run_source`).
- Adds **WD8** (rbwasm build pipeline) and **WD9** (in-browser transport).
- Hosting reframe: serving the wasm as a **static asset** (the try.ruby-lang.org model) sidesteps the ~1 MB Cloudflare **Workers** cap entirely — it never applied to Pages/static CDNs.
- **Does not flip the production default** — the Rack backend (Option B) stays the deployment until the build resolves WD6 ② and ③.

**2. Adds the build scaffolding** under `plugins/rigor-playground/wasm/`
| File | Role |
| --- | --- |
| `Gemfile` | The playground's exact runtime set (`rigortype` carries `rigor-rbs-inline` via require_paths; + `rbs-inline`; + `ruby_wasm`). |
| `Rakefile` | Two-step `build` (`rbwasm build` → `rbwasm pack`) + `serve`/`smoke`/`clean`. |
| `vfs/boot.rb` | In-VM adapter — mirrors the backend's `Rigor::CLI.start` invocation so the WD2 JSON is byte-identical. Carries a WASI `flock`/`fsync` shim confined to the wasm boot; the engine is untouched. |
| `vfs/.rigor.yml` | Packed config (loads `rigor-rbs-inline`, strict). |
| `index.html` | Static frontend; reuses the CodeMirror UI, swaps `fetch()` for `vm.eval`. |
| `README.md` | Build/run steps, prerequisites, WD6 status, fallbacks. |

## Verification

- **Adapter is contract-correct independent of wasm.** Loaded against the real engine in the Flake, `boot.rb` reproduces the backend's diagnostics for the sample — including the inline-RBS `call.argument-type-mismatch` for `:bad` against `:asc | :desc`. This confirms config discovery + `rigor-rbs-inline` loading + JSON shape; only the wasm *compile* is unproven by this.
- **Lint + syntax clean** on the new Ruby; the wasm `Gemfile` resolves and installs (`prism`/`rbs` native exts build).

## Outstanding — the WD6 ② spike

The remaining unknown is purely whether `rbs`'s C extension links under wasi-sdk. The `rbwasm` from-source build (Ruby 4.0 + `prism` + `rbs`) is **running at PR-open time** (currently compiling deps with wasi-sdk-24.0). I'll update this PR with the result:
- **completes** → WD6 ② green, `rake serve` gives a fully in-browser playground;
- **`rbs` fails to link** → that's the concrete blocker to report upstream, and Option B stays the answer.

Build artifacts (`build/`, `vendor/`, the `.wasm`) are gitignored; this PR is source-only.


zonuexe:
## Build-spike result (WD6 ②) — ran end-to-end

The `rbwasm` from-source build was run to completion. Outcome:

**✅ `prism` + `rbs` link cleanly** under `wasi-sdk-24.0` — zero linker errors. This is the binding question the ADR shelved Option A on (the `rbs` C-extension). It's cleared.

**❌ Build aborts on `psych`** (stdlib YAML):
```
wasm-ld: error: ext/psych/psych.a(...): undefined symbol: yaml_get_version
wasm-ld: error: ext/psych/psych.a(...): undefined symbol: yaml_emitter_initialize
…
```
`libyaml` itself cross-compiles fine — `ruby_wasm`'s `LibYAMLProduct` produces `libyaml.a` and `configure` gets `--with-libyaml-dir` pointing at it (the same path that links `openssl`/`zlib`). But `psych`'s `-lyaml` doesn't reach the final static `wasm-ld` link. This is a **`ruby_wasm` static-ext link-wiring gap, not an `rbs`/Rigor problem** — and Rigor needs `psych` (parses `.rigor.yml` + the `data/builtins` YAML catalogs), so it can't be excluded.

### Net
The ADR's headline risk is gone; what's left is a narrower, known-shaped ruby.wasm packaging issue. Next steps (cheapest first), now recorded in the ADR WD6 ② and `wasm/README.md`:
1. bump `ruby_wasm` when a release past 2.9.4 ships (possible version-specific regression);
2. patch the crossruby link to add `libyaml.a` to `XLDFLAGS` explicitly;
3. report upstream to `ruby/ruby.wasm` with the trace.

Production default is unchanged — Option B (Rack backend) stays until a runnable `.wasm` lands. Pushed as `ae916472`.

zonuexe:
## Runtime verified — the playground actually runs in the browser VM (WD6 ③)

After the green build, I ran the wasm headlessly in a **browser-representative** VM (`@ruby/wasm-wasi` + `browser_wasi_shim`, writable `/` preopen — exactly what the page's `DefaultRubyVM` provides) and it now produces **diagnostics byte-identical to the Rack backend**:

```
L2:3   [call.undefined-method]       undefined method `upcase' for nil
L10:21 [call.argument-type-mismatch] expected :asc | :desc, got :bad   (← rigor-rbs-inline)
```

Getting there took four wasm-runtime integration fixes (each found via `wasm/smoke.mjs`, now wired to `rake smoke`):

1. **`gem "js"`** — `DefaultRubyVM` won't instantiate without it (the `JS.global` bridge).
2. **`require "rubygems"`** — ruby.wasm runs rubygems-disabled (`Gem` is a stub; `/bundle/setup` only unshifts `$LOAD_PATH`). RBS's loader needs real rubygems; without it the type universe built **empty** (0 classes → nothing flagged). Now 334+ classes load.
3. **`/bundle/setup`** instead of `bundler/setup` (which hits the disabled rubygems path).
4. **Writable `/work`** — WASI mounts the packed tree read-only; the adapter stages a writable working dir (the browser VM's in-memory `/`) for the cwd-relative cache + per-request buffer.

`rake smoke` asserts the two expected diagnostics and exits non-zero otherwise.

### State
- ✅ Build green (prism + rbs + psych link; libyaml patch)
- ✅ Runtime correct in wasm (WD6 ③), ~70 MB → R2 confirmed (WD10)
- ⏳ Remaining: R2 bucket + secrets (enable the `playground-wasm` publish step), then the docs-site `public/playground/` sync + `<meta>` R2-URL injection.

Pushed `41fc46f4`. Production default still unchanged (Option B Rack backend) until the R2 + site wiring lands.

PR #62, opened 2026-07-10
Seed module declarations into the cross-file discovery index (ADR-57 WD3)

zonuexe:
Closes the module-singleton facade slice ADR-57 tier 4 left open.

## What was wrong

`ScopeIndexer#record_declarations` has always registered `module` and `class` declarations
identically, but the project-wide seed `collect_class_decls` registered only classes. So a module
constant read from a **sibling file** typed `Dynamic[top]` while a class typed `Singleton[C]` —
and the module facade (`Feature.enabled?`, `Gitlab::Utils.to_boolean`) is exactly how applications
consume these modules. `Feature.enabled?` alone is 695 unprotected GitLab sites.

The seed's exclusion comment justified itself with a fall-through hazard (`Kernel.select ->
Array[String]`) and deferred "until cross-file `discovered_methods` follows the same project-wide
seed". Both halves have expired:

- The cross-file `singleton_def_nodes` index landed with ADR-24 slice 2.
- The hazard does not reproduce. Kernel's private instance methods resolve to nothing on a
  `Singleton[M]` receiver. Of the 28 Object/Kernel/Module/Class methods that *do* resolve on a bare
  `module Empty; end`, 26 are correct for a module object. The two that are not (`M.new`,
  `M.superclass`) mistype only code that raises `NoMethodError` at runtime — and they pre-date this
  change, since the per-file pass already registers modules.

Design note: `docs/notes/20260710-module-singleton-cross-file-seed.md`. Decision: ADR-57 WD3.

## Payoff

GitLab `lib`, both sides re-measured, identical 93,796-site denominator:

| | protected | ratio |
|---|---:|---:|
| before | 23,044 | 0.2457 |
| after | 25,856 | **0.2757** |

**+3.00 pp / +2,812 protected sites.** For scale, the whole ADR-67 call-site parameter-inference
lever measured +0.75 pp on Mastodon.

## Gate (ADR-57 WD2 — adjudicate per firing class, fix artifacts at root)

`make verify` clean, `make bench-perf` clean (28.52 M allocations, ceiling 29.17 M), `make docs-check`
clean. Corpus diff, `check --no-cache --no-baseline`:

| corpus | before | after |
|---|---:|---:|
| haml `lib` | 59 | 59 |
| kramdown `lib` | 68 | 68 |
| liquid `lib` | 2 | 2 |
| rgl `lib` | 70 | 70 |
| Mastodon `app/models` | 5 | 5 |
| GitLab `app` | 284 | 284 |
| Redmine `app`+`lib` | 72 | **73** |

Two firings adjudicated:

**rigor's own `lib` (2 errors) — fixed at root.** `CLI::DiagnosticFormats.render` is a `case/when`
with no `else`, so its now-inferred return is `String | nil`, while both call sites gate on
`.supports?` and immediately call `output.empty?`. The adopted type was right and the invariant was
real but unencoded. `render` now raises on an unrecognised format, which also turns a
`FORMATS`/dispatch drift into a loud failure instead of a nil every caller must guard.

**Redmine (+1 `possible nil receiver`) — an independent artifact class, not caused by this change.**
`Redmine::CodesetUtil.replace_invalid_utf8` now resolves cross-file to its honest `String | nil`; the
call site excludes the nil arm only through an ActiveSupport `login.present?` guard.
`Narrowing#resolve_rbs_extended_method` reads `rigor:v1:predicate-if-true` facts only for a
`Nominal` / `Singleton` receiver, so **a union receiver never receives predicate facts at all** —
reproduced in six lines, and confirmed unfixed by adding the annotation. Any precision work that
turns a `Dynamic` into `T | nil` under a `present?` guard surfaces it.

Queued as its own FP-removal slice with its own corpus gate, because a new narrowing rule carries a
new false-positive envelope. Its general form needs no annotation: on the truthy edge of a zero-arg,
block-less predicate over a union receiver, drop every arm whose RBS return type is literally `false`
(`NilClass#present?: () -> false` already says so); mirror on the falsey edge.

## Left as future slices

- The `Singleton[Object]` fallback receiver for modules (the `M.new` / `M.superclass` leak) — it
  mistypes only code that raises, so it is precision hygiene, not FP work.
- Singleton-ancestry resolution (`extend SomeModule`, inherited class methods) — already named a
  future slice by ADR-57 tier 4.
- Registering `extend self` bodies into `singleton_def_nodes` — the instance-def consult in
  `try_user_method_inference` already resolves them.


zonuexe:
Rebased onto #63 (the union-arm predicate polarity slice), per the front-load discipline — remove the
false positive before adding the precision that exposes it.

With #63 in place, this PR's one Redmine firing is gone at its root, and the corpus delta is **zero
everywhere**:

| corpus | base (#63) | this PR |
|---|---:|---:|
| haml / kramdown / liquid / rgl `lib` | 59 / 68 / 2 / 70 | identical |
| Mastodon `app/models` | 5 | 5 |
| GitLab `app` | 268 | 268 |
| Redmine `app`+`lib` | 69 | **69** |

`make verify` clean on the rebased branch. The base should be retargeted to `master` once #63 merges.

PR #65, opened 2026-07-10
Label an unresolved constant owned by an RBS-less locked gem (ADR-82 WD9)

zonuexe:
GitLab plan P2-7, recorded as [ADR-82](docs/adr/82-dynamic-provenance-wiring.md) WD9. Closes the
`external_gem_without_rbs = 0` half of ADR-82's G2: two corpora reproduce that the one
user-actionable tractability bucket fires on zero sites while GitLab's lockfile holds 806 gems with
no RBS — so `coverage --protection` calls 99.4% of holes `engine_gap` when many have the answer a
user can act on.

## The plan's framing was mechanically impossible

The plan said: *"when a dispatch's receiver class is owned by an RBS-less locked gem, record the
external-gem cause."* But a no-RBS gem's receiver never carries the class name — the constant read
itself (`Faraday`) fails resolution and widens to `Dynamic[top]` with the generic
`unsupported_syntax` cause at `fallback_for`. By dispatch time the name is gone. (The two dispatcher
tiers that do record this cause require ADR-10 / `pre_eval:` opt-ins — exactly why the bucket
measured zero.)

The honest recording site is the **constant-resolution miss**
(`ExpressionTyper#unresolved_constant_fallback`), and ADR-82 WD6's chain inheritance then carries the
cause through `Faraday.new.get(...)` with no further work.

## Ownership by reading, never guessing

For each locked gem classified `:missing` by `RbsCoverageReport`:

1. Resolve its installed source root from RubyGems **metadata** (`Gem::Specification#full_gem_path`)
   — no gem code loads, the ADR-72 "loads data only" posture.
2. Prism-parse its conventional entry file (`lib/<name>.rb`, dash → directory variant) and index the
   **top-level** class / module / constant declarations under their root name.

A camelize-the-gem-name heuristic is deliberately rejected — it breaks exactly where it matters
(`activesupport` → `ActiveSupport`) and is the guessing the ADR-82 honesty criterion forbids. The
constants come from the parse; the entry-file path is the require-name convention `Bundler.require`
itself depends on.

**Everything fails open.** Uninstalled gem, absent or unparseable entry file, a root constant
declared only in a deeper file, a project typo (`Farraday`) — all keep the generic cause. The failure
mode is a missing label, never a wrong one, which is also what makes the bounded entry-file scan
acceptable: coverage can grow later (full-tree scan behind a per-gem-version cache) with no soundness
question.

The index is built **lazily** on the first unresolved constant — a project whose constants all
resolve never pays — and the scan is entry-file-bounded, so paying it is sub-second at 806-gem scale.

Design note: `docs/notes/20260711-external-gem-constant-provenance.md`.

## Gate

Precision-additive per the ADR-75 side-channel contract — no type, diagnostic, or severity change:

- `make verify` clean (853 files rubocop clean; index unit specs + scanner-level chain/fail-open
  specs; the `Environment#missing_rbs_gem_owner` addition is pinned in the public-API drift snapshot
  and `sig/rigor/environment.rbs`).
- `make bench-perf` green (28.64 M allocations, ceiling 29.17 M).
- Redmine `app`+`lib` diagnostics byte-identical (69 → 69).

**Re-bucketing measurement (the ADR-82 WD5-style functional gate), Mastodon `app/models`, identical
denominator:**

| cause | before | after |
|---|---:|---:|
| `unsupported_syntax` | 2,259 | 2,212 |
| `external_gem_without_rbs` | 0 | **47** |
| `none` / `inferred_return_untyped` / `explicit_untyped` | 923 / 824 / 41 | byte-stable |

Every sampled labeled site adjudicated **correct**: all root at `I18n` (`I18n.t` in validations,
`I18n.locale`, `I18n.available_locales`) — the i18n gem is locked, ships no RBS, and its RBS exists
in `gem_rbs_collection`, so the `add_rbs` routing is genuinely actionable. The `app/models` yield is
small by construction (models are AR-dominated and plugin-typed; gem facades live in services/lib) —
the GitLab `lib` measurement is running and will be posted on this PR.

## Known imprecision, accepted and recorded

A gem declaring a generic top-level constant (`module Util`) can claim a reference the project meant
for its own `Util` in an **unanalyzed** path (analyzed declarations win earlier, at
`discovered_classes`). Accepted: the label is a side-channel hint that can fire no diagnostic, the
two constants genuinely collide at runtime, and the alternative (refusing generic-looking names) is
itself a heuristic.


zonuexe:
## Resolver correction + honest coverage story (pushed)

While adjudicating the GitLab `lib` measurement (124 sites), I found the first cut's gem-directory
resolution was subtly wrong, and the fix clarifies exactly what this feature covers.

**The bug:** the first cut resolved a `:missing` gem's source dir via
`Gem::Specification.find_by_name`. But rigor runs under its **own** bundle
(`BUNDLE_GEMFILE=<rigor>/Gemfile`), so that sees rigor's gems, not the target's — it resolved only
the handful of gems both bundles share (`i18n`, `rack`, `activesupport`) and missed every
project-specific gem. The measured 47/124 were entirely that shared subset, correctly attributed but
a floor.

**The fix:** the primary resolver is now the target's bundle install tree
(`<bundle>/ruby/*/gems/<name>-<version>/`, the pure-filesystem layout `BundleSigDiscovery` already
walks), threaded through `for_project`. `Gem::Specification` stays a fallback for a no-bundle project
(sound — a gem's top-level namespace constant is version-stable).

**What this means for coverage** (now stated plainly in the note + WD9):

| target install layout | coverage |
|---|---|
| `vendor/bundle` / `BUNDLE_PATH` (Docker/CI norm) | **full** external-gem population — the case the fix unlocks |
| global gems (mise/rbenv, both survey corpora) | only gems rigor itself bundles, until target-`GEM_PATH` awareness lands |

Both survey corpora happen to be global-gems installs, so their measured yield (Mastodon `app/models`
47, GitLab `lib` 124) is the shared-gem floor — sound, but not the ceiling a `vendor/bundle` run
reaches. Closing the global-gems gap needs rigor to learn the target's `GEM_PATH`, the *same*
limitation `BundleSigDiscovery` already has; recorded as a shared follow-up. Fail-open throughout, so
the withheld coverage is a missing label, never a wrong one.

`make verify` + `make bench-perf` + `make docs-check` green; no survey regression.

Commits: `7436ea42` (resolver), on top of `30fdc180` / `de72f11e`.

PR #76, opened 2026-07-12
Allocation-free AST child iteration

zonuexe:
## Allocation-free AST child iteration

`Prism::Node#compact_child_nodes` allocates a fresh `Array` on every call — for the ~43 leaf node classes (`IntegerNode`, `LocalVariableReadNode`, `NilNode`, …) it is literally `def compact_child_nodes; []; end`. Rigor's tree walkers call it unconditionally on every node of every walk, so one full-tree walk allocates one throwaway array per node visited. On leaf-heavy sources (mail's Ragel-generated parsers have hundreds of thousands of integer-literal leaves, and the engine walks the tree many times) these arrays are **over half of all allocations** in a run; on Rails corpora and kramdown they are ~10–19%.

### What changed

Loading `Rigor::Source::NodeChildren` compiles a **`#rigor_each_child` method onto every Prism node class** (additive reopening of a dependency class — the `Cache::RbsEnvironmentMarshalPatch` precedent; the `rigor_` prefix keeps it collision-free) that yields the same children, in the same order, without materialising the array. Each generated statement mirrors `compact_child_nodes`' own form for that field kind exactly: required node fields yield unconditionally, optional fields behind a truthiness guard, list fields iterate the stored array in place. That makes iteration **one virtual send per node with zero allocation** — the same dispatch mechanism `compact_child_nodes` itself uses. The field map is derived once at load from `Prism::Reflection`, so it tracks the installed `prism` (ADR-79).

**Dispatch variant history** (why a method on the node class): a `public_send`-per-field loop measured ~45% *slower* than `compact_child_nodes` on a full-tree walk (dynamic dispatch per field swamps the allocation win), and a central generated-method table dispatched via `Module#send` + per-class Hash lookup was wall-neutral on the interpreter but still cost ~5% on Rails-corpus walks under YJIT. The compiled node-method form beats or matches `compact_child_nodes` everywhere measured.

Every unconditional `node.compact_child_nodes.each { … }` walker in `lib/` is migrated to `node.rigor_each_child { … }`: the shared node walker, the engine-owned plugin / check-rule node walk, the `ScopeIndexer` discovery-seed and cross-file pre-pass descents, the five check-rule collectors, the inference walkers, the sig-gen collectors, the dependency-source walker, the protection mutator, the node locator, the annotate command, and the five language-server providers. Traversal order, per-walker child-type guards, and `break` / `next` / `return` semantics are preserved exactly, so **diagnostics are byte-identical**. A nil-tolerant `NodeChildren.each_child(node)` wrapper remains the public entry for callers holding a maybe-nil slot.

Bundled `plugins/*/lib` walkers are intentionally left for a separate slice; only the engine-owned plugin node walk was migrated. A project `sig/prism_node_children.rbs` declares the reopening so Rigor's own self-check resolves the method statically — its teeth flagged all 51 call sites as `undefined-method` until the runtime definition was declared to the RBS universe, which is the discipline working as designed.

### Equivalence contract

`spec/rigor/source/node_children_spec.rb` asserts, over the analyzer's own source plus an exotic-syntax corpus spanning ≥130 Prism node classes, that **both** `#rigor_each_child` and `each_child` are element-for-element identical (object identity + order) to `compact_child_nodes` for every node reached, and that every concrete node class carries the compiled method.

### A/B — interpreter (cold `--no-cache`, order-alternating interleaved pairs, best-of)

| target | allocs master→branch | wall master→branch | diagnostics |
| --- | --- | --- | --- |
| mail (`lib`) | 21,025,502 → 9,197,438 (**−56.3%**) | 4.84s → 4.37s (**−9.8%**) | 26 = 26 ✓ |
| kramdown (`lib`) | 4,150,403 → 3,362,009 (**−19.0%**) | 1.63s → 1.59s (−2.5%) | 68 = 68 ✓ |
| liquid (`lib`) | 4,556,749 → 4,112,243 (**−9.8%**) | 1.44s → 1.40s (−2.8%) | 5 = 5 ✓ |
| mastodon (`app/models`) | 10,092,492 → 8,554,210 (**−15.2%**) | 4.16s → 3.86s (**−7.2%**) | 267 = 267 ✓ |
| redmine (`app/models`) | 13,709,303 → 11,270,595 (**−17.8%**) | 6.87s → 6.74s (−2.0%) | 55 = 55 ✓ |

Allocations drop on every target and wall improves on every target. `mail` + `mastodon-models` diagnostics JSON arrays are byte-identical between engines (diffed, not just counted).

### A/B — YJIT (`RUBY_YJIT_ENABLE=1`, cold)

| target | allocs master→branch | wall | verdict |
| --- | --- | --- | --- |
| mail (`lib`), 2 pairs | 21,025,849 → 9,197,790 (**−56.3%**) | 4.04s → 3.93s (**−2.7%**) | faster |
| mastodon (`app lib`), 6 order-alternating pairs | 55,314,301 → 46,171,117 (**−16.5%**) | mean 14.90s → 14.84s; median 14.70s → 14.81s; per-pair deltas +10.1/−0.2/+1.2/−5.3/+2.3/−8.1% → **pair-mean −0.00%** | match (noise-dominated; master's own spread 13.5–16.9s) |

The mastodon "app lib" YJIT probe — the flagship criterion — shows a paired-delta mean of −0.00% with the branch mean (14.84s) below master's (14.90s) and inside the reference 14.87–15.20s band; diagnostics identical (2348) on all 12 runs.

### Gates

- Equivalence spec green (both dispatch layers, ≥130 node classes).
- `make verify` green (full suite; rubocop clean; self-check + `check-plugins` clean).
- Corpus diagnostics byte-identical vs master on all five targets (counts match; mail + mastodon JSON diffed).
- `make bench-perf` green — rigor's own `lib` allocations 29.3M → 15.1M (−48%).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,11 @@ Older release notes are archived under [`docs/`](docs/) when the leading version
 
 ## [Unreleased]
 
+### Changed
+
+- **[engine]** AST tree walks no longer allocate a throwaway array per node visited, cutting total allocations on every `rigor check` run — most sharply on leaf-heavy sources.
+  - `Prism::Node#compact_child_nodes` builds a fresh `Array` on every call, and Rigor's walkers called it unconditionally on every node of every walk. On a Ragel-generated parser (mail's `lib`, with hundreds of thousands of integer-literal leaf nodes) those arrays were over half of all allocations. A new `Rigor::Source::NodeChildren.each_child` yields the same children in the same order without the array — reading each child field directly and reusing list fields' stored arrays — and the engine's tree walkers (the shared node walker, the plugin/check-rule walk, the scope-discovery seed pass, the collectors, and the sig-gen / dependency / mutation scanners) now use it. Diagnostics are byte-identical; the field map is derived from `Prism::Reflection` so it tracks the installed `prism`.
+
 ## [0.2.9] - 2026-07-11
 
 v0.2.9 sharpens Rigor on large Rails applications, with GitLab-scale projects as the proving ground: PostgreSQL `db/structure.sql` is accepted as a schema source, the strong-parameters chain stays typed, several route-helper and ActiveSupport coverage gaps close, and module facades now resolve across files. Protection-coverage tooling gets faster with a fork-parallel scan and more actionable — it now names when a missing-RBS gem is the cause of a hole ([ADR-82](docs/adr/82-dynamic-provenance-wiring.md)). Other fixes strengthen control-flow narrowing around `present?` / `blank?` guards and in-body mutation, and make the persistent cache robust across upgrades.
--- a/lib/rigor/analysis/check_rules/always_truthy_condition_collector.rb
+++ b/lib/rigor/analysis/check_rules/always_truthy_condition_collector.rb
@@ -2,6 +2,8 @@
 
 require "prism"
 
+require_relative "../../source/node_children"
+
 module Rigor
   module Analysis
     module CheckRules
@@ -81,7 +83,7 @@ module Rigor
           visit(node) if conditional_node?(node) && !in_loop_or_block
 
           child_in_loop_or_block = in_loop_or_block || enters_loop_or_block?(node)
-          node.compact_child_nodes.each { |child| walk(child, in_loop_or_block: child_in_loop_or_block) }
+          Source::NodeChildren.each_child(node) { |child| walk(child, in_loop_or_block: child_in_loop_or_block) }
         end
 
         def conditional_node?(node)
--- a/lib/rigor/analysis/check_rules/dead_assignment_collector.rb
+++ b/lib/rigor/analysis/check_rules/dead_assignment_collector.rb
@@ -2,6 +2,8 @@
 
 require "prism"
 
+require_relative "../../source/node_children"
+
 module Rigor
   module Analysis
     module CheckRules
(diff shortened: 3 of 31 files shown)


zonuexe:
**Follow-up landed: `0e80d04e` — Compile `rigor_each_child` onto the Prism node classes.**

Post-merge interleaved re-measurement under YJIT showed this PR's central dispatch table (`Module#send` + per-class Hash lookup per node visit) still regressing mastodon `app`+`lib` cold wall ~5% despite the allocation win. The follow-up replaces it with a method compiled onto each Prism node class (field reads inlined from the same `Prism::Reflection`-derived table, statement-for-statement identical to `compact_child_nodes`' own forms) — one virtual send per node, zero allocation.

A/B vs this PR's merge commit (`ef4f345e`), cold, order-alternating interleaved pairs:

| target | wall | allocs | diagnostics |
| --- | --- | --- | --- |
| mail `lib`, interpreter, 2 pairs | 5.00s → 4.36s mean (**−12.7%**) | equal | 26 = 26, JSON byte-identical |
| mastodon `app lib`, `RUBY_YJIT_ENABLE=1`, 3 pairs | 15.90s → 15.04s mean (**−5.3%** pair-mean; all pairs negative) | equal | 2348 = 2348 |

vs the pre-#76 engine (`0a4adf48`): every interpreter target now faster (mail −9.8% wall at −56% allocs, mastodon `app/models` −7.2% at −15%), and the mastodon `app lib` YJIT probe is a wash across six order-alternating pairs (paired-delta mean −0.00%, 14.84s vs 14.90s mean).

Also in the follow-up: a project `sig/prism_node_children.rbs` declaring the reopening — the self-check's teeth flagged all 51 call sites as `undefined-method` until the runtime definition was declared to the RBS universe; the equivalence spec now covers both the compiled instance method and the `each_child` wrapper.

Gates at push: `make verify` green (7591 examples, lint, self-check, `check-plugins`), `make bench-perf` green, corpus byte-identical.

Note: the commit was intended as a PR from `perf/node-child-method-dispatch` but landed on `master` directly — the branch was cut with `origin/master` as upstream and the push refspec expanded to it. All gates were green before the push; happy to revert to PR flow if preferred.

PR #93, opened 2026-07-14
Enable binpacker work stealing on the CI profile

zonuexe:
## Problem

[CI run on PR #86](https://github.com/rigortype/rigor/actions/runs/29257276571/job/86840811973?pr=86) showed a 65s / 15.4% max worker deviation across the 4 binpacker workers:

```
Worker 0: 75 tests, 7m44.3s
Worker 1: 78 tests, 5m55.1s
Worker 2: 79 tests, 7m02.9s
Worker 3: 80 tests, 7m17.5s
Balance: max deviation 1m04.8s (15.4%)
```

Root causes (measured by replaying our own `tmp/binpacker.timings` — 591k lines, ~81 runs of cached history):

1. binpacker's per-file weights summed the **entire append-only history**, over-weighting long-lived files ~N× and starving new specs — 7.2% deviation from this alone even with otherwise perfect predictions. Fixed upstream in [rigortype/binpacker#12](https://github.com/rigortype/binpacker/pull/12).
2. CI run-to-run noise (per-file CV ~0.13 weight-weighted) puts a **~6% deviation floor under any static schedule** — no weight quality fixes this; only dynamic rebalancing does.

## Change

Enable `steal_enabled: true` on the ci profile: a worker that finishes its LPT queue pulls batches from the queue with the most remaining work. Costs a few extra rspec boots per worker; simulation on our real weights (4 workers, 3s/boot) shows ~2% deviation and the best makespan of the policies tried, and stealing is robust even under the pre-fix weights (0.6–1.6%), so this doesn't need to wait for the binpacker release. Once binpacker 0.4.0 (with weight-guided batches + bounded timing file) is released and the lock bumped, boot count drops further (~5/worker) and the CI timing cache shrinks from 51MB-scale history to ~3 samples per test.

Also drops the dead `strategy: static` key binpacker never read.

Expected CI effect: wall time for the spec job ≈ −30–40s and Balance line at a few percent; this PR's own CI run is the validation.

Local `default` profile stays static (unchanged behaviour).

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -25,7 +25,7 @@ GEM
     ast (2.4.3)
     base64 (0.3.0)
     bigdecimal (4.1.2)
-    binpacker (0.3.0)
+    binpacker (0.4.0)
     concurrent-ruby (1.3.7)
     connection_pool (3.0.2)
     diff-lcs (1.6.2)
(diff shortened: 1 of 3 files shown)


zonuexe:
**Blocked on the next binpacker release** (carrying [rigortype/binpacker#12](https://github.com/rigortype/binpacker/pull/12)).

The first CI run of this PR deadlocked and was cancelled: with binpacker 0.3.0, dynamic batches let the spawned rspec process inherit the orchestrator's control pipe as stdin, which stays open between batches — so `spec/rigor/cli_spec.rb`'s "rigor lsp returns 0 when stdin closes cleanly" example blocked forever reading LSP frames (static mode masked this because the pipe closes right after the single batch is sent). Fixed at the root in binpacker (worker stdin now points at /dev/null; regression spec included).

Merge order: release binpacker with #12 → `bundle update binpacker` here → re-run CI on this PR.

zonuexe:
binpacker 0.4.0 is released and the lock is bumped — this is no longer blocked.

Verified locally with the released gem under the `ci` profile: all 7708 examples pass across 12 workers, no deadlock, and the reported totals are accurate. Handing the rest to CI on 4 workers, where the Balance line is the number to watch (was 15.4% / 65s).

PR #100, opened 2026-07-15
Frame ADR-50 WD5 as an edition versioning model, not strict semver

zonuexe:
Resolves the ADR-50 WD5 support-line question by naming the versioning model it was already implicitly using, and correcting a couple of stale docs alongside.

## The problem

WD5 described "support" as a release count — *latest + previous minor* — and pivoted vaguely to "the `1.x` branch" post-1.0. That reads as **strict semver**. But WD1/WD3 already make a *minor* carry behavioural change: new diagnostics land off-by-default behind `bleeding_edge:` or are absorbed by the baseline. That is the **PHP / PHPStan edition model**, not semver — and leaving the model unnamed made the support rule look like an arbitrary count rather than a consequence.

It also surfaced a real question: post-1.0, is there any demand to backport to a *previous minor* (`1.2.x` after `1.3.0`)? Under the edition model the answer is **no** — a minor upgrade is soft by construction (bleeding-edge + baseline absorb it), so there is nothing to be stranded behind. The only hard break is the **major**, so the only line worth a maintenance branch is the previous *major*.

## What changes (docs only)

**ADR-50 WD5** is rewritten to:
- **Name the model.** Patch = fixes; **minor = soft break** (features/diagnostics, always upgradable, at most a baseline refresh — PHP `8.4`/`8.5`, PHPStan `2.1`/`2.2`); **major = hard break** (frozen-surface change, bleeding-edge graduations default-on, deprecations removed — PHP `8.0`/`9.0`).
- **Split "support" into two commitments.** Troubleshooting/answering is always-on and version-agnostic — not a per-line promise. A *maintenance line* is the narrower thing: security + correctness backports, **BC-safe only** (never a feature or a new diagnostic — a project on the line chose stability).
- **Tie the support line to the hard break.** A maintenance line exists only for the release behind the most recent *hard* break: pre-1.0 the previous minor (`0.y-1`), post-1.0 the previous major (`(N).x` after `(N+1).0`). No per-minor line within a major.
- **Make it demand-driven and explicitly not guaranteed** — matching the comparator: PHPStan publishes no LTS window and winds a previous line down once the successor is the default (`2.1.x` became a "final send-off" after `2.2.x`).

**`docs/compatibility.md`** trajectory section rewritten to match, and its stale "the overlay is empty today" corrected (`reject-unparseable-signatures` shipped in #96).

**`docs/ROADMAP.md`**: the ADR-50-remaining summary follows the new framing, and two stale `flow_contribution_for` bullets are fixed — ADR-52 deleted the hook and migrated all five consumers, so that is not open work.

## Verification

`make docs-check` clean. No code change.


zonuexe:
Folded in the **branch model** (WD5 addition, `8fe49967`): single-trunk with lazy tag-rooted maintenance branches, weighed against the language (PHP/Ruby long-lived cut branch) and PHPStan (branch-per-minor, moving default) comparators. A maintenance branch is cut from the previous line's release tag only when a backport is actually needed — a tag is a permanent branch point, so eager per-hard-break branches would carry lines with no users, and a per-minor branch is machinery for the per-minor patching WD5 declined. `docs/compatibility.md` gets the one-line mechanics. `make docs-check` clean.

PR #112, opened 2026-07-16
Normative status fidelity: the founding-era stratum and the declare-or-mark gate

zonuexe:
## What this is

`CLAUDE.md` promises the spec binds, and rigor-rs is the reader that cannot fall back on `lib/` — for the port a normative clause *is* the requirement. Probing that promise found **three independent places where the corpus states behaviour that never shipped**, none recorded as deferred anywhere:

| # | Clause | Implementation |
| --- | --- | --- |
| 1 | `special-types.md` § `void` — two MUSTs ("keeps `void` distinct internally", value-context diagnostic, generic-slot preservation) | `rbs_type_translator.rb:51` — `Bases::Void => :translate_untyped`. No carrier, no rule id, zero specs |
| 2 | `diagnostic-policy.md`'s 12-family taxonomy | `static.*` / `compat.*` / `hint.*` / `generated.*` have **zero** implemented ids |
| 3 | `internal-type-api.md`'s "contract every type object MUST satisfy" | `normalize` / `traverse` / `consistent_with` / `equal_value` / `has_method` / `subtype_of` absent from all 23 carriers |

All three are founding-era (ADR-1/2/3) design targets presented as binding contract. **ADR-1 even foresaw #1** — "`void` and `untyped` are likely to be treated as broad aliases too early" — and it happened, unobserved.

**Why no gate caught them.** Every `manual_drift_spec.rb` docs axis runs impl → doc; the reverse is unchecked. And every analysis gate is FP-oriented, so a diagnostic that was never implemented is *silence* — invisible by construction. (ADR-62 kinship: mutation testing needs code to break; an absent feature has none.)

## The decision (ADR-92)

> A normative clause states behaviour in the present tense only if it ships. Otherwise it carries an explicit status marker.

The force is the corollary: **silence is never the honest state** — marking is always available and always cheap, so the cost of implementing or narrowing never justifies a clause that lies. This separates the status question (settles now) from the design question (waits for evidence). It is **ADR-49 axis 6 applied to the spec corpus** instead of the ADR corpus; `inference-budgets.md` and the `sig.*` row already demonstrate the marker, so only enforcement was missing.

## What lands

- **Markers** (WD2/WD3) for `void`, the four empty families, and `internal-type-api.md`'s method surface (narrowed to the routing that ships).
- **The gate** (WD4) — axis 5 of `manual_drift_spec.rb`, the doc → impl direction. Reads the **full 39-id vocabulary** (`ALL_RULES`'s 26 + the 13 non-check ids), since `ALL_RULES` alone reports false gaps. Load-bearing both ways: removing a marker from an unimplemented family goes red; leaving a marker on a family that later ships goes red.
- **The `void` measurement**, resolving part of the carry-over.

## The void finding worth reading

The engine matches **neither the spec nor RBS**:

| | `void` is… |
| --- | --- |
| RBS (`docs/syntax.md`) | `top` — "They are all equivalent for the type system; they are all *top type*" |
| The spec (intent, unshipped) | distinct; diagnoses value use; materializes as `top` |
| The engine | `untyped` = `Dynamic[top]` — **looser than both** |

So the engine carries a silent divergence from RBS that ADR-79's fidelity criterion does not license; it survived because the direction is FP-safe.

**Measured**: patching `Bases::Void => :translate_top` is byte-identical on mail (26) / kramdown (68) / haml (60) / liquid (5) / mastodon `app/models` (0, 248 files) — non-vacuously, since instrumenting the translator counts **29 real `Void` translations in kramdown**, 2 in mail. Narrowing to RBS is **free**, which eliminates the concede-to-the-engine option.

Deliberately **not** implemented here: the choice between narrowing to RBS and implementing the distinct-`void` intent is a design call, recorded with its evidence for a follow-up.

## Scope

**Docs + one spec axis. `lib/` is untouched** (`git diff master -- lib/` is empty) — the translator patch was measurement-only and reverted.

## Verification

`make docs-check` green (246 examples), `rubocop spec/docs/` clean, `git diff --check` clean.

## Grounding

`docs/notes/20260716-dspec-formal-spec-substrate-evaluation.md` — evaluates mizchi/dspec as a formal-spec substrate (rejected: its Clause AST is a propositional fragment over uninterpreted atoms) and Lean 4 / Alloy as a CI oracle (rejected: the formalizable layer isn't the buggy layer, and a model with no conformance link to `lib/` reproduces the `void` failure one level up). A PBT spike over the shipped type algebra found no law violations across 2,000 cases / 13 carriers — the independent evidence for that verdict. Its real yield was that *writing* the laws is what surfaced finding #3.


zonuexe:
Scope grew by one finding — the sharpest of the set.

**A fourth instance, and a different class**: not silent non-implementation but a **direct contradiction between the binding spec and an accepted, shipped ADR**. `overview.md` § "Compatibility hierarchy" (2026-04-28) requires inline rbs-inline annotations to be "always parsed and used whenever present" and says Rigor "**MUST NOT require `# rbs_inline: enabled`** to begin parsing them". ADR-32 (2026-05-25, shipped v0.1.10) made ingestion an opt-in plugin whose WD2 default requires exactly that magic comment — and lists the spec-mandated behaviour as a *rejected alternative*, citing eight other ADRs but never the month-older clause it contradicts. Per CLAUDE.md the spec binds → the shipped default is non-conforming on both the activation model and the gate.

Reconciliation is cheap: upstream honours `# rbs_inline: disabled` **unconditionally** (`parser.rb:73`, before the `opt_in` branch), so the plugin's existing `require_magic_comment: false` mode (ADR-32 WD10) is exactly the spec's semantics. What diverges is the default and the wiring, not the engineering.

Landed in `03ed1294`:
- the marker in `overview.md` at the point of declaration (ADR-92 WD3 shape)
- an ADR-92 amendment recording the fourth instance — notably, the WD4 gate *cannot* catch this class (it reads the diagnostic-family table; prose stays ungated by WD1), so the manual probe remains the instrument for prose clauses
- **ADR-93 (proposed, measurement-gated)**: flip `require_magic_comment:` to `false`, default-wire the bundled plugin presence-gated on the library resolving (ADR-72's on-disk gate shape + ADR-90's bundle fallback) — a recorded partial reversal of the ADR-27/31 auto-load deferral for one bundled plugin — with the standalone residual (no library anywhere) deferred to the WD4 corpus measurement. Nothing implemented; `lib/` still carries only the `void → top` translation change.

`make docs-check` green (247), whitespace clean.

zonuexe:
ADR-93's WD4 measurement ran (`31a69a26`) — it found a blocker, fixed it, and reshaped the proposal. The PR now carries a second `lib/` change beyond `void → top`.

## Finding 1 — a blocker, and a live bug for today's opt-in users

Running the rbs-inline plugin on **herb** (the one survey project with pervasive real annotations *and* a hand-written `sig/` over the same code) **collapsed the entire RBS env**: 1,490 classes → 0, `require` itself stopped resolving, 74 false `call.unresolved-toplevel`. `sig/` + inline overlap is the *expected* state for a migrating project — every opt-in user with both hits this today, independent of ADR-93.

Mechanics (not where the rescue assumed): `RBS::Environment#add_source` appends to `sources` **before** inserting decls, so the mid-insert `DuplicatedDeclarationError` was caught per-entry but the **poisoned source stayed behind**, and `resolve_type_names` — which rebuilds from `sources` — re-raised outside every rescue. Fixed: the skip is transactional, the explicit `.rbs` wins, a resolve-time backstop covers the rbs `>= 3.0, < 5.0` range, and the drop is loud on both paths (once-per-run banner + `virtual_rbs_collision_quarantined`, derived from the built env so a **cache HIT reports it too**).

Two methodology traps recorded in the note: my first reading ("RBS defers duplicate detection to resolve") was wrong, and my first fix (a retry loop) failed silently on herb because its `unload` re-raised *inside* the rescue handler — re-verifying the minimal-repro green against the real corpus is what caught it.

## Finding 2 — post-fix herb A/B/C is sane

| mode | diagnostics | delta |
|---|---|---|
| A: no plugin | 11 | — |
| B: opt-in, magic honored (today) | 11 | ±0 (both magic files quarantine cleanly against `sig/`) |
| C: all-inline (ADR-93 target) | 12 | **−3 genuine wins**, +4 pre-existing `Regexp.last_match(1)` imprecision unmasked (group always participates → runtime nil impossible; `sig/`'s `-> untyped` had masked it) |

## Finding 3 — the naive always-parse wiring is refuted, WD1/WD2 reshaped

On **mail** (zero annotations), all-inline mode moved 26 → 42: upstream's opt-out mode synthesizes a **full `-> untyped` skeleton for every unannotated def**, and accepted signatures outrank body inference — the skeleton *fights* Rigor's inference-first analysis. The spec binds Rigor to honour *annotations* whenever present, not to manufacture untyped shadows of unannotated code. ADR-93's conforming default is therefore **annotation-presence-gated** synthesis (no magic comment required — the MUST NOT holds), never on-for-everything.

## Verification

Full suite **7,920 / 0 failures** (+6 loader specs: collision / sig-wins / parse-vs-collision / warn-once; the existing sig-vs-stdlib degrade specs still pass). `make check` / `check-plugins` / lint / docs-check (247) clean. **No-plugin path byte-identical on mail** — the fix is inert without the plugin.

PR #113, opened 2026-07-16
Gate the magic-comment-free rbs-inline mode on annotation presence (ADR-93 WD1)

zonuexe:
Follow-up to #112. ADR-93's WD4 measurement **refuted the ADR's own WD1 twice** — which is the whole reason the gate existed.

## Refutation 1 — the mode fought the analysis

`require_magic_comment: false` mapped straight onto upstream's opt-out mode, which synthesizes a full `def f: (untyped x) -> untyped` skeleton for **every unannotated def**. Rigor trusts an accepted signature over body inference, so those skeletons *replaced* real inferred types. On **mail** (zero annotations): **26 → 42** diagnostics.

Upstream isn't wrong to emit them — `rbs-inline --output` exists to **generate** a `sig/`, so a skeleton is the right answer to *its* question. The lesson: delegating the **grammar** (ADR-32 WD3) does not mean delegating **what counts as a contribution**. The spec binds Rigor to honour annotations *whenever present*; it does not ask for untyped shadows of unannotated code.

## Refutation 2 — `#:nodoc:`, found because the fix fell short

Gating on annotation presence only got mail to **31**. RDoc directives collide lexically with `#: <type>`, and upstream reads `class Foo #:nodoc:` as a type assertion of an alias named `nodoc` (consumes the word, drops the trailing colon). It's one of the most common comments in Ruby — **61 of mail's files** opted in on that alone. The mis-parse is upstream's and worth reporting there, but it's **harmless in the output** (renders back as a plain `# :nodoc:` comment, not a bogus type), so the filter belongs at the gate only.

Detection delegates to upstream's own `AnnotationParser`, not a regexp — the grammar is upstream's, and a regexp would opt in a doc comment containing a URL like `http://example.com#:`.

## Result

| corpus | base | gated | |
|---|---|---|---|
| mail (no annotations) | 26 | **26** | 42 → 31 → 26 |
| kramdown / haml / liquid | 68 / 60 / 5 | same | no-op ✓ |
| herb (real annotations) | 11 | 12 | **−3 genuine wins** / +4 |

The **default mode is unchanged** (`require_magic_comment: true`), so no configured project moves. This mode's only current users are the browser playground and `--treat-all-as-inline-rbs`; the playground's sample still synthesizes, and a pasted *unannotated* snippet now keeps its inference instead of being skeletonized.

## Also in this PR

ADR-93's WD1/WD2/WD4 rewritten to the measured shape — the previous text described a design the measurement had already refuted, with the refutation recorded only in a later section (a status-fidelity defect of exactly the class ADR-92 addresses). **New WD1a**: the herb +4 are `Regexp.last_match(1)` possible-nil, adjudicated as pre-existing imprecision that `sig/`'s `-> untyped` had masked — so per the ADR-57 protocol, match-success narrowing is a **prerequisite of the default flip**, not a follow-up.

**Recorded residual**: within an annotated file, an unannotated def still gets a skeleton. Filtering per member would couple to upstream's AST (rejected by ADR-32 WD1/WD3 on grammar-drift grounds) and the measured cost on herb is nil.

## Verification

Full suite **7,925 / 0 failures** (+6 plugin specs). `make check-plugins` / lint / docs-check (247) clean, `git diff --check` clean.


zonuexe:
`e5bea6a6` — filtering `#:nodoc:` at the gate turned out to be half a fix, and my reasoning for stopping there was wrong.

I had checked the directive only on a **class**, where upstream renders it back as a `# :nodoc:` comment, and concluded it was harmless in the output. In a **trailing** position upstream emits the directive name **as the type**:

```ruby
def f(x) #:nodoc:
```
```rbs
def f: (untyped x) -> nodoc
```

Nothing declares `nodoc`, so `DefinitionBuilder` raises `NoTypeFoundError` **for the whole class** and every real annotation in it is dropped without a warning. Measured:

| | `size_of`, which carries `#: (String) -> Integer` |
|---|---|
| no `#:nodoc:` in the class | `Integer` |
| one sibling gains `def internal(x) #:nodoc:` | `2` — the annotation is gone |

`stub_missing_referenced_types` does not catch it: that tier takes `project_sig_files`, so a virtual buffer's undeclared references never reach it.

**Fix**: rewrite each directive to its spaced spelling before synthesis (`# :nodoc:`, which upstream's grammar ignores). Matching on shape rather than a name list covers all 17 directives the [Ruby docs](https://docs.ruby-lang.org/ja/latest/library/rdoc.html) list, and none of `#: () -> void`, `#: String`, `#:Integer`, `#:my_alias`, `#: { name: String }`, `#[Integer]` matches it — a valid RBS type never opens with a bare lowercase word closed by a colon. Prism decides what is a comment, so a directive-shaped string literal stays put.

Scale: rbs-inline emits 29 of these for Ruby's own [`lib/fileutils.rb`](https://github.com/ruby/ruby/blob/15c71c929a25e6c8c1b2bb8ff3edd41b99b7868b/lib/fileutils.rb#L186) — 49 across 8 first-party files in ruby/ruby, plus 128 more in vendored copies. Reported upstream as [soutaro/rbs-inline#248](https://github.com/soutaro/rbs-inline/issues/248).

**One trap, now pinned by a spec**: `Prism::Location#start_offset` counts **bytes** while `String#insert` indexes **characters**. My first cut put the space mid-word (`#:n odoc:`) on any file with multi-byte content and left the directive live — mail's own `field.rb` caught it, at 26 → 32 diagnostics. `start_character_offset` is the fix.

Corpus after: mail 26, kramdown 68, haml 60, liquid 5, all byte-identical; herb keeps −3 wins. Full suite **7,930 / 0 failures** (17 plugin specs), lint / check-plugins / docs-check clean.

zonuexe:
`bd9af404` — **ADR-94** records the rbs-inline support situation and the rbs 4.x question, per the discussion. No code; the migration is deferred.

## rbs 4.0 absorbed the reader

soutaro called the `rbs-inline` gem a prototype in [soutaro/rbs-inline#9](https://github.com/soutaro/rbs-inline/issues/9) — *"the implementation will be merged to rbs-gem"* — and it happened. rbs **4.0.0** ships `RBS::InlineParser` + `RBS::Source::Ruby`; rbs **3.9.4** ships none of it.

That dissolves ADR-32 WD1's premise: it rejected a core reader because `rbs-inline` would become a core runtime dependency against ADR-0, but **`rbs` already is one**. A reader inside `rbs` would also retire ADR-93's WD2 (default-wiring) and WD3 (the standalone residual, which exists *only* because a bare install has no `rbs-inline` gem).

## Measured: both of this PR's workarounds fix defects rbs's parser lacks

| source | rbs-inline 0.14.0 | `RBS::InlineParser` |
|---|---|---|
| no annotation | `def f: (untyped x) -> untyped` | `(?) -> untyped` |
| a real annotation | honoured | honoured |
| `def f #:nodoc:` | `-> nodoc`, a type nothing declares | `AnnotationSyntaxError`, method stays untyped |
| annotation + `#:nodoc:` in one class | the class fails to build, every annotation lost | the annotation survives |

**This corrected my account of the skeleton problem.** The harm was never the `untyped` *return*: `(untyped x)` declares an arity and a parameter type, which outranks inference (`"A"` → `Dynamic[top]`), while `(?)` (`RBS::Types::UntypedFunction`) claims nothing and leaves inference alone. That is why mail moved 26 → 42 under rbs-inline's opt-out mode and would not under rbs's.

## Decision: deferred

> A dependency migration that buys correctness a workaround already buys does not justify narrowing the supported toolchain range.

The migration's whole cost is the **rbs 3.x floor** (ADR-79's `>= 3.0, < 5.0`, kept honest by #22's `Gemfile.rbs-compat` matrix). Its benefit is deleting ~40 lines of plugin code that works. Dropping 3.x is not planned for v0.3.0 or near it, so the reader follows the floor rather than forcing it.

## Two things stay separate

**A live bug, found while measuring**: a `(?)` method type in a hand-written `.rbs` crashes `rigor check` **today** — `NoMethodError: undefined method 'required_positionals' for an instance of RBS::Types::UntypedFunction`. `RbsDispatch` guards the form for blocks; the `CheckRules` arity path does not. Reachable with no migration (rbs 4.x core RBS may adopt `(?)`), and a prerequisite of one. Lands on its own (ADR-94 WD2).

**Coverage is unmeasured** (WD4): four probed cases against the grammar ADR-32 inherits for free (generics, mixin generics, `@rbs inherits`/`override`, block-introduced types, attributes, ivars, constants, `@rbs!`, `%a{…}`). Migrating on an unmeasured subset would trade the workarounds for silent precision loss.

`make docs-check` 248 green.

PR #119, opened 2026-07-16
Docs economy + development-flow re-org: ADR indexes, contract import, issue-tracked backlog (ADR-97/98)

zonuexe:
Rigor's two ADR lists had both stopped doing their job. Each carried a dense per-ADR essay restating the ADR body — **~256 KB of restatement**, independently worded, and beginning to contradict the bodies.

| | at branch point | now |
| --- | --- | --- |
| `CLAUDE.md` (loaded every session) | 134,688 B — 89% ADR bullets, ~34k tokens | **16,310 B** — a 10-entry premise set |
| `docs/adr/README.md` (read on demand) | 141,169 B — status cells avg 1,420, peak 5,195 chars | **17,055 B** — status only, and its table renders again |

## The finding

**This audit already happened.** `db8d01bf` (2026-05-29) found the `CLAUDE.md` defect in nearly the same words and applied the same fix, 69,380 → 15,542 bytes. Seven weeks later it was back at **8.7×**.

**In both files the declared contract was already correct — only the entries drifted, and at the same ADR number.** `CLAUDE.md`'s header calls it "a navigation index"; the README's "How to Read" says the Status column is `Accepted`/`Proposed`/`Superseded` plus a parenthetical.

| ADR range | CLAUDE.md bullet | README status cell |
| --- | --- | --- |
| ADR-0 … 39 | 116 B avg | 19 chars median |
| ADR-40 … 96 | 2,029 B avg | 585 → 3,089 chars, rising by decade |

Entries below ADR-40 are still compliant, verbatim, in both files. Nothing fattened — **every ADR authored after ADR-40 entered at the then-current density**. One ratchet, two files, two mechanisms, neither a discipline failure: `rigor-adr-author` § 4c said *"keep the one-liner consistent in density with its neighbours"* (a ratchet, with no absolute reference), and `rigor-prior-art` listed the bullets as its **"First stop"** (so the density had a real consumer).

So the thing worth acting on is not length; it is that **compressing by hand does not hold**.

## Is compressing the README lossy?

The cells are worded independently of the bodies — median 4-gram overlap only **30.6%**. That measures wording, not substance, so **eight cells were adjudicated by reading each against its full ADR body**: **six are pure RESTATEMENT**; ADR-89 and ADR-49 hold one unique fragment each, **migrated into their bodies first** — which is what makes this lossless rather than merely smaller.

The adjudication also found the cells failing the *other* way: **ADR-48's said `Struct` was deferred when its body records slices 1–3 landed**, and **ADR-73's dropped two dated amendments**, so a cell-only reader would never learn `rigor skill --full` exists. The 58 rewritten cells are therefore derived from each ADR's own `Status:` block, which ADR-92 already makes canonical.

## Not an index — a premise set

The first cut kept all 98 in `CLAUDE.md` at ≤ 100 chars each. That answered the token question but not the premise question, and review caught it. A flat list asserts every ADR is equally worth a session's attention, when 88 are **lookups** — reached only by a session already in their area.

> The test is not "is this ADR important?" but **"will a session that never thought to look it up do the wrong thing?"**

ADR-5's robustness principle silently governs every type anyone authors here. ADR-93's rbs-inline default flips govern only a session already editing rbs-inline ingestion. So `CLAUDE.md` now carries **only the premises** — the foundation and conceptual core (ADR-0–5) plus the standing policies in force (ADR-31, 49, 50, 97) — **10 entries, capped at 12**. The line is the README's own: its "How to Read" already names ADR-0 the foundation, ADR-1–3 the conceptual core, and everything above readable *"as needed"*.

The cap matters more than its value: a new ADR now normally adds **nothing** here, and an eleventh premise costs a deliberate argument. That is a stronger ratchet-stopper than a per-entry budget — it removes the default of "append mine too" rather than bounding it. `rigor-adr-author` § 4c becomes *"usually nothing to do"*.

## A rendering bug the compression exposed

The README's table carried **five blank lines between rows**. A blank line ends a markdown table, so the index rendered as six tables, each promoting the next ADR row to a header. **This predates the branch** (confirmed at the same five positions on `origin/master`) and survived precisely because nobody reads a table rendered when every cell is a thousand-character essay. It was obvious within minutes of the cells getting short — criterion 2 arriving from a second direction, so the gate grows an axis for it.

## What ADR-97 decides

Criterion 1 — **an index entry's payload does not vary with the indexed document, so it has a fixed budget** (which is also what reconciles the caps with ADR-49's rejection of absolute length limits: that governs ADR *bodies*, whose payload varies with stakes). Criterion 2 — **an economy rule with no mechanical gate is a temporary state, not a decision.**

- **WD1** — `CLAUDE.md`: a premise set, ≤ 12 entries, ≤ 100-char topics.
- **WD2** — `docs/adr/README.md`: the complete index; status only, ≤ 200 chars, derived from the body's Status block.
- **WD3** — `rigor-prior-art` repoints at the **ADR bodies**. Costs it little: its actual method is ripgrep over `docs/adr/*.md`; an index's job there is to shortlist.
- **WD4** — all of it gated in `spec/docs/agent_index_spec.rb` under the existing `make docs-check`.

The gate's progress-vocabulary check is deliberately narrow: `deferred`/`rejected` are **not** progress words, because for an evaluation ADR the deferral *is* the decision (ADR-95, ADR-86). A gate that fires on a correct entry teaches authors to route around it.

Deferred, recorded rather than silent: the ADR **bodies**' own economy — ADR-49's rule, proportional to stakes, a separate pass.

## Verification

- `make verify` — exit 0, 13 suites green. The `doctor_command.rb:279` warning in its output is **pre-existing on master** (confirmed against a stashed clean tree); no `lib/` code is touched here.
- `make docs-check` — 263 examples, 0 failures. `rubocop` clean; `git diff --check` clean.
- **Gate teeth verified on every axis.** Re-inflating ADR-58's bullet fails the topic cap + status-word axes; restoring ADR-82's old cell fails the status cap + status-word axes; opening a cell with "Landed and working" fails the status-word axis; re-inserting a blank table row is caught at `line 102: ""` (the exact pre-existing bug); appending three non-premise ADRs fails on the cap (`13 entries (cap 12)`), ordering, and slug parity. The set-parity axis caught the missing README row for ADR-97 while it was being authored.

No `CHANGELOG.md` entry: contributor-facing docs + a dev gate, not user-observable (AGENTS.md § CHANGELOG Style).


zonuexe:
Fifth commit (`d4b28c46`) extends the arc from *document economy* to the *development flow itself* (ADR-98):

- **The backlog is now GitHub Issues** — 45 issues (#120–#164) migrated after per-item adjudication (four independent passes against code / ADR status blocks / git history; ~40% of "open" ROADMAP items turned out to be shipped). Triage vocabulary (`needs-triage` / `needs-info` / `ready-for-agent` / `ready-for-human` / `wontfix`) + `area:*` labels + `v0.3.0` / `v1.0.0` Milestones.
- **`docs/ROADMAP.md` deleted** — strategy was ADR-50's, milestones were CHANGELOG's, the backlog is issues'. Frozen records keep their "ROADMAP §" citations as history; live docs rewired.
- **`docs/CURRENT_WORK.md` is a 62-line full-replace handoff**; its gotchas moved to the owning skills/ADRs, its open items are issues.
- **Engineering-skills config** scaffolded per `setup-matt-pocock-skills`: `docs/agents/*` + root `CONTEXT.md` glossary; external PRs are a triage surface.
- **Gates**: ROADMAP must not exist; handoff ≤ 120 lines. Teeth verified (both fail correctly when mutated).

`make verify` exit 0, `make docs-check` 265 examples green, rubocop clean.

PR #223, opened 2026-07-26
Fix broken documentation links in README

f440:
The rigor.typedduck.fail site no longer serves the `/reference/` path segment for manual, handbook, type-specification, and ADR pages. Links pointing at those URLs 404 in production.

This removes `/reference/` from nine documentation links and retires the roadmap link — the static page has been replaced by the GitHub Milestones view, which is what the project actually uses for release planning (ADR-50, ADR-98).


zonuexe:
@f440 Thank you!

Issue #152, opened 2026-07-17
Narrowing: widen the `&&`/`||` polarity gate beyond Constant-only


zonuexe:
The pre-1.0 architecture re-examination (Phase 3, `docs/notes/20260610-lib-rigor-architecture-rereview.md`) converged certainty judgments onto `Narrowing` as sole owner (`predicate_certainty` / `class_pattern_certainty` / `value_pattern_certainty`), with Typer and Evaluator as callers. One fragment deliberately stayed out of that convergence: `constant_value_polarity` in `lib/rigor/inference/expression_typer.rb` (the `&&`/`||` value-position polarity gate) only handles `Constant` types today. Extending it to a full-probe (any type, not just literals) would improve `&&`/`||` narrowing precision, but is explicitly flagged as a real behaviour change (not a pure refactor) that needs its own evaluation of FP risk before landing — queued as demand-gated for that reason, not because it's low-value.

_Migrated from docs/ROADMAP.md § "Internal architecture — pre-1.0 re-examination" (dissolved by ADR-98)._

zonuexe:
> *This was generated by AI during triage.*

## FP evaluation — measured. Verdict: do not land the widening.

This issue was demand-gated on an evaluation nobody had run. It has been run (2026-07-17). **Zero measurable benefit, one demonstrated harm mechanism.**

### The widening is smaller than it looks

`Type::Combinator.union` drops `Bot` members, so today's fallback already computes the exact short-circuit result on the hand-off-to-right edges: `a && b` with truthy `a` gives `union(Bot, b) = b`. The entire widening reduces to **dropping the statically-dead right operand** on two edges: `a || b` with certainly-truthy `a`, and `a && b` with certainly-falsey `a`.

`||=` is **not** in the blast radius — the `*OrWriteNode`s route to separate handlers. The memoization idiom is structurally safe.

### Measurements (prototype: swap the gate for `Narrowing.predicate_certainty`; all runs inside the Flake)

| Measurement | Result |
| --- | --- |
| `make check` self-check | diagnostic-identical to baseline |
| Full spec suite | 7976 examples, 0 failures — **nothing pins the changed edges** |
| 14-target corpus diff (kramdown, net-ssh, liquid, mail, faraday, rubocop-ast, rgl, algorithms, haml, slim, parser, jbuilder, kdtree-book, mastodon app/models; 363 baseline diagnostics) | **zero drift** |
| Instrumented census of the two new edges, same corpus | **1 firing** |

[...]

PR #292, opened 2026-08-06
Stop the branch elision from resting on an optimistic carrier

### LINKED ISSUE #286: if/unless truthiness drops an arm on nominals nil and false conform to (Object, Kernel)
## Parent

Found during #152's FP evaluation (`docs/notes/20260805-issue-152-and-or-polarity-gate-fp-evaluation.md`). #152 was declined precisely because widening `constant_value_polarity` would let the engine bet that a non-Constant value is provably truthy — which `docs/internal-spec/inference-engine.md` (line ~251) forbids by a normative MUST. The finding here is that **the `if` / `unless` path already makes that same bet today, unguarded.**

## The behaviour

On current master:

```ruby
x = Time.now
y = if x then 1 else "s" end   # y types as `1`, not `1 | "s"`
```

The `else` arm is dropped because the condition's type is judged truthy. And the judgement reaches the shapes #152's evaluation classified as outright unsound: `Nominal[Object]`, `Nominal[BasicObject]`, `Nominal[Kernel]`, `Nominal[Comparable]` all answer `:truthy`, even though `nil.is_a?(Object)` and `false.is_a?(Kernel)` are both true (`falsey_nominal?` matches exact class names only, so `nil` and `false` slip through any supertype spelling).

For `Time.now` the bet happens to be correct. For a variable typed `Object` — which is what an unannotated parameter or a permissive RBS return frequently gives — it is not.

## Why it matters

The reasoning that declined #152 applies here verbatim: a value union or nominal that omits `nil` is **optimistic, not proof** (the spec's own words about non-static-key `Hash#[]`), so concluding truthiness from it discards an author-written branch. In the `&&`/`||` position the cost is a deleted fallback; here it is a deleted `else` arm — the same class of harm, already shipped.

This is not a claim that the current behaviour is causing measured false positives — #152's corpus diff found none attributable to it. It is a consistency and soundness question raised by that evaluation: one path is forbidden from making a bet the other makes freely, and the spec text only mentions the `&&`/`||` gate.

## What to decide

1. **Measure first.** How often does the `if`/`unless` truthiness judgement actually drop an arm on the corpus, and how many of those rest on an unsound nominal (`Object` / `BasicObject` / `Kernel` / `Comparable`) rather than a genuinely non-falsey shape? Instrument, run over the survey targets with `--no-baseline --no-cache` (and see #285 — a warm cache will otherwise serve pre-edit results and fabricate a zero).
2. Then choose: tighten `falsey_nominal?` to treat any nominal that `nil`/`false` conform to as undecidable (the narrow, clearly-correct fix); or extend the spec's exclusion to name this path too and state deliberately why the bet is acceptable here.

## Acceptance criteria

- [ ] The corpus measurement above, recorded here.
- [ ] A decision, with the spec text updated in the same commit if the behaviour is kept (`docs/internal-spec/inference-engine.md` currently constrains only the `&&`/`||` gate).
- [ ] If tightened: zero new diagnostics on `lib`, `plugins/*/lib examples/*/lib`, and the corpus.
- [ ] `make verify` exits 0.

zonuexe:
Closes #286.

## What this changes

Rigor deliberately reads past core RBS's `%a{implicitly-returns-nil}` — `Hash#[]` types as `V`, `Array#first` as `E` — because pessimising them to `V?` costs 25 measured false positives on our own `lib`. `docs/internal-spec/inference-engine.md` records that such a value is **optimistic, not proof**, and forbids `flow.always-truthy-condition` and the `&&`/`||` gate from concluding truthiness from it.

There is a third reader of that judgment and the spec never named it. The `if`/`unless` branch elision consumes `predicate_certainty` too, and unlike the other two it is not `Constant`-gated — so it makes the forbidden bet freely. The consumers also differ in what the bet costs: the gate only withholds a diagnostic, while the elision **deletes the branch the program takes when the lookup misses**. On master today:

```ruby
MAP = { a: "x", b: "y" }.freeze
v = MAP[key]
n = if v then 1 else "none" end   # else arm dropped; n types as 1
n.upcase                          # error: undefined method `upcase' for 1 — the program is correct
```

## The criterion (ADR-101)

A certainty judgment may rest on a value's nil-freeness only when that nil-freeness is a property of the value — its class, its inhabitance, its literal — never when it is a property of our own modelling choice. The value stays fully usable for dispatch: "we chose this" is a fact about the *judgment*, not a defect in the *type*.

Three properties are load-bearing:

- **Per selected overload.** `Array#first` is marked, `Array#first(3)` is not, and a signature that already spells the miss as `?` (`String#[]`, `Enumerable#find`) is never marked.
- **Not applied to a shape-tier read.** A static key on a declared shape, or `first` on a non-empty `Tuple`, proves the read cannot miss.
- **Declined at the consumers, never in `Narrowing`.** `&&=` / `||=` and the and/or surviving-left edge read `falsey_nominal?` / `narrow_falsey` too; widening the falsey fragment there re-admits `nil` into a bound local and buys `possible nil receiver` firings — a soundness fix paid for in FPs.

The mark rides the ADR-75 / ADR-82 side channel rather than any type carrier, and is excluded from `Scope#==` / `#hash`, so it cannot vary a flow decision.

## Why not the cheaper gate

"Decline for non-`Constant` carriers" was implemented, measured, and rejected. It is wrong in both directions at once. It **over-declines**: only 35 of the 134 non-`Constant` verdicts are actually optimistic; the rest are nominals whose class genuinely excludes `nil`, plus `Tuple`/`HashShape` carriers sound by inhabitance. It **under-declines**: a `Hash` whose values share one type yields a lone `Constant` that is exactly as optimistic as the union case, so 12 optimistic verdicts survive it — eight of them one redmine cluster where the verdict is `:falsey` and the elision drops the arm that actually runs.

## Measurement

Eleven projects (Rigor `lib`, plugins+examples, erubi, faraday, net-ssh, kramdown, liquid, mail, textbringer, redmine, mastodon), `--no-cache --no-baseline --no-ci-detect --workers 0`:

- 47 of 2,060 verdicts affected, 30 of them dropping a written arm.
- **Diagnostics byte-identical on every target, in both directions** — zero added, zero removed.
- Precision does not regress: Rigor's own `lib` *gains* 3 `constant` nodes and loses 3 `bot (unreachable)` ones, because the previously-skipped arm now gets typed. redmine moves 50.93% → 50.92%.

A zero diff is not on its own evidence — #152 passed a "zero new diagnostics" gate and shipping it would still have been wrong, because its damage mode was deleted fallbacks rather than new firings. The damage mode of *declining* is lost precision, so precision was measured alongside, and the A/B flag was independently proven live on both a small target and a large one before the zero was accepted.

Full method and per-target numbers: `docs/notes/20260806-issue-286-optimistic-carrier-provenance-census.md` (already on master). The measurement harness is preserved on `optimistic-nil-free-provenance-census-286`.

## Tests

`spec/rigor/inference/optimistic_origin_spec.rb` pairs every decline example with a proof-side control that must still elide — a decline assertion is the easiest test in this repo to pass by accident, since any construction error also widens the type. Both mutants were run: "always decline" kills the six controls, "never decline" kills the six decline examples.

`make verify` passes; `make check` and `make check-plugins` stay clean.


zonuexe:
Cross-checked the whole supported `rbs` range (`>= 3.0, < 5.0`) rather than just the pinned bundle, per the lesson from the last version-ranged dependency: `TypeDef#overload_annotations` exists on both lines and `Array#first` / `Array#last` / `Hash#[]` all carry `implicitly-returns-nil` under RBS 3.10.4 as well as 4.0.2, so the fix is effective across the declared range and not only on the development pin.

That check also caught a latent trap in the spec, now fixed. Two examples inferred their carrier from `ENV`, whose core signature differs between the two RBS lines, so they failed under 3.x for a reason unrelated to the behaviour under test — and CI would never have shown it, because the `rbs-compat` job only runs `spec/rigor/environment`. Those fixtures now bind the carrier directly (`Array[String]`), which keeps the load-bearing pair — `xs.first` declining and `xs` still eliding, the same carrier class separated only by provenance — independent of RBS version. All 14 examples pass on both lines, and both mutants still kill six examples each.

zonuexe:
Re-ran the corpus A/B against the **shipped** build, because the measurement in the PR description does not cover it. The census build propagated the mark through locals only; this branch also propagates through instance variables, so it declines strictly more verdicts than what was measured, and the earlier zero could not be carried forward on its own.

Re-measured `master` vs this branch across all eleven targets (`--no-cache --no-baseline --no-ci-detect --workers 0`, cwd = target): **still byte-identical, zero added and zero removed** — rigor `lib` 0/0, plugins+examples 0/0, erubi 3/3, faraday 6/6, net-ssh 23/23, kramdown 68/68, liquid 2/2, mail 25/25, textbringer 184/184, redmine 69/69, mastodon 23/23.

Auditing the diff also raised an ordering question the spec did not answer: the value-side gate consults the mark *before* typing the predicate, so `if MAP[key]` — the read as the predicate itself, with no intervening binding — could in principle be judged before the mark exists. It is covered, because the statement evaluator types the predicate first and only then asks for the verdict, but nothing held that in place; every other example routes through a binding and so exercises the name-keyed tables. Pinned in 0f7ba8a1.

Issue #299, opened 2026-08-07
Resolve the bundled-RBS definition-build collisions surfaced by #295's warning


zonuexe:
> *This was generated by AI during triage.*

## What to build

Follow-up to #295: the definition-build-failure warning that landed in [#296](https://github.com/rigortype/rigor/pull/296) immediately surfaced 7 classes in Rigor's OWN bundled signature set whose `RBS::DefinitionBuilder` build fails, so every call on them silently degraded to `Dynamic[top]` until now (and still degrades — the warning only made it visible). Fixing the collisions themselves was declared out of scope for #295; this issue is that work.

Captured on master `ce994b6e` (`rigor check lib` plus a two-line probe touching `BigMath` / `Nokogiri::CSS::Parser`; paths relativized, pins: rbs 4.1.1, bigdecimal 4.1.2):

```
Gem::Specification       DuplicatedMethodDefinitionError  rbs-4.1.1/sig/shims/rubygems.rbs #version  vs data/vendored_gem_sigs/rubygems/rubygems_extras.rbs:114
Bundler                  DuplicatedMethodDefinitionError  rbs-4.1.1/sig/shims/bundler.rbs .default_lockfile  vs data/vendored_gem_sigs/bundler/bundler.rbs:81
Bundler::LockfileParser  DuplicatedMethodDefinitionError  rbs-4.1.1/sig/shims/bundler.rbs #initialize  vs data/vendored_gem_sigs/bundler/bundler.rbs:137
Time                     DuplicatedMethodDefinitionError  rbs-4.1.1/core/time.rbs #utc?  vs data/gem_overlay/activesupport/core_ext.rbs:281
DateTime                 DuplicatedMethodDefinitionError  rbs-4.1.1/stdlib/date/0/date_time.rbs #to_time  vs data/gem_overlay/activesupport/core_ext.rbs:460
BigMath                  DuplicatedMethodDefinitionError  bigdecimal-4.1.2/sig/big_math.rbs .E  vs rbs-4.1.1/stdlib/bigdecimal-math/0/big_math.rbs:24
Nokogiri::CSS::Parser    NoSuperclassFoundError           data/vendored_gem_sigs/nokogiri/nokogiri.rbs:80 — Could not find super class: Racc::Parser
```

Four distinct root causes:

[...]

PR #302, opened 2026-08-07
Bump up version to 0.3.2

zonuexe:
Seals `[Unreleased]` as `## [0.3.2] - 2026-08-08` and moves `Rigor::VERSION`, `Gemfile.lock` and the README status line with it.

v0.3.2 makes the mutation tier of `rigor coverage` practical on a whole project (per-file cache, `--workers` fork parallelism, and two opt-in bleeding-edge features that give it `rigor check`'s cross-file knowledge), widens the editor surface to whole-project scope on save / in-flight buffers / multi-buffer publishing, types a dry-schema Contract's own result shape, and retires a batch of engine false positives. The largest correctness fix is the least visible: Rigor's own bundled signatures collided with the ones the `rbs` gem ships, silently disabling type checking for `Gem::Specification`, `Time`, `Bundler` and six more classes.

## Changelog sealing

The section had accumulated 44 bullets across 53 merged PRs, and landing-time discipline had slipped on two axes: only 4 bullets carried the PR link the release-prep contract requires (0.3.0 and 0.3.1 both ran at 100%), and about half the top-level bullets had grown a second clause behind an em-dash or a comma splice. Every one of the 43 sealed bullets is now a single sentence carrying its landing PR plus its reporting issue, with the "why" and the measured numbers moved into child items.

Consolidations that change what a reader sees:

- The two #299 entries collapse into one bullet about nine classes that silently lost type checking, with the stdlib-extension gating (#301) as a child item.
- The dry-schema bullet that carried two unrelated features behind an "and" splits into the nested-`each` recursion and the `dry-schema.unknown-type` diagnostic.
- The Tier-2 kill oracle (#262), the LSP pool's batch-size gate (#281), the `URI.extract` fold (#284) and the incremental-snapshot key (#291) are now linked on the child items they belong to.

Fifteen merged PRs are deliberately unlinked — perf-baseline recalibrations, changelog and self-check gates, self-mutation spec work, and two dependabot lockfile bumps. None is user-facing.

The archival rule does not fire: 0.3.2's `(0, 3)` prefix already matches the most recent archive boundary, which moved `0.2.x` out at 0.3.1.

## Verification

- `make verify` green (8,899 examples, RuboCop clean, `check` + `check-plugins` clean)
- `make docs-check` green (325 examples)
- `git diff --check` clean
- `gem build rigortype.gemspec` builds `rigortype-0.3.2.gem` (not committed)


zonuexe:
Advisory perf gate: wall-only failure, explained and not recalibrated.

Three attempts on this branch gave `wall_s` 14.23 / 11.75 / 12.08 against an 11.0 limit, while both deterministic signals stayed inside their bands every time (`allocations` 16,426,325 ≤ 16,557,991; `peak_rss_kb` 280,520–304,440 ≤ 332,090).

Measured the same two commits on one local machine to separate runner noise from a real regression:

| | wall_s | allocations |
| --- | --- | --- |
| baseline commit 58b42c74 (2026-07-30, the calibration point) | 8.107 / 8.286 | 15,871,199 |
| this branch | 8.552 / 8.257 | 16,529,199 |

Wall is flat on identical hardware, so the +30% is a property of the Linux runner, not of the code. Per the release-prep contract, wall alone is never a reason to recalibrate `bench/baseline.json`, so the baseline is untouched.

One thing the run does surface, inside the band rather than past it: `allocations` grew 15,871,199 → 16,529,199 (+4.1%) over this cycle, deterministic and reproducible both locally and on CI. The +5% band permits 16,557,991, so the headroom is now about 29k allocations — roughly 0.2%. The band keeps its teeth only while the baseline tracks the real cost, so the next allocation-adding change trips this gate. Worth a refresh from a CI-measured Linux run on master after this release lands, or an issue if the growth has an owner worth finding first.

Issue #347, opened 2026-08-13
Reverse index of resolved constant references, and a minimal rigor unused report


zonuexe:
## Parent

#344

## What to build

The tracer bullet: a thin but complete path from constant-reference recording through to a rendered report. After this slice, `rigor unused <path>` runs on Rigor's own `lib` and on a corpus Rails app and prints a candidate list.

End to end, the slice covers:

- **Recording.** A reverse index of *resolved* constant references, built by a **dedicated pass over constant-read nodes**, resolved through the same lexical rules as `Reflection.lexical_constant_candidates`. Modelled on `Analysis::DependencyRecorder` (ADR-46 slice 1) for its discipline — off by default, cheap disabled path, purely observational, never changes a diagnostic — but **not** riding the typing path. See the correction below.
- **Reachability.** Mark-and-sweep from a root set to the project's declared class/module constants, restricted to declarations whose source lies inside the analysis root. Roots in this slice come from one place only: a config-declared glob of entry-point paths. The plugin-supplied root mechanism is a later slice, and the report must be honest that it is missing.

### Correction from the #345 spike — do not ride `resolve_constant_type`

This issue originally named `Reflection.resolve_constant_type` as the recording point. **Measurement showed that is insufficient**, because that method fires only where the engine needs a constant's *type*; a constant that is read but never typed leaves no trace. On Rigor's own `lib` the resulting index produced 140 candidates of which **zero** were genuinely unused. The under-counting is in at least five distinct places: cross-file value constants (`in_source_constants` is per-file by design), parameter-default expressions, lambda/proc bodies that are constant rvalues, superclass positions, and intermediate namespace segments of a `ConstantPathNode`. Details and evidence in #345.

Two further constraints the spike established:

- **Declaration ownership must mean "declared *first* here", not "declared here".** Reopening a foreign class currently registers it as a project declaration — `RBS::Location`, reopened by the marshal patch, was a class-tier candidate for exactly this reason. Neither `discovered_classes` nor `discovered_class_sources` distinguishes a definition from a reopening today.
- **Class/module constants only at launch.** The value-constant tier cannot work until `in_source_constants` crosses file boundaries, which is an engine change with its own FP surface and is out of scope here.
- **CLI.** A `rigor unused` subcommand in the `rigor triage` / `rigor coverage` family — a renderer plus `--format json`, not a diagnostic emitter.
[...]

Issue #345, opened 2026-08-13
Spike: measure the false-positive rate of unused-constant candidates on Rails corpora


zonuexe:
## Parent

#344

## What to build

A throwaway measurement harness that answers, with numbers from real Rails corpora, the question #344 turns on: **how bad is the naive "constants with zero resolved references" list, and how much does each root source actually buy?**

The article's decay curve on one proprietary app was 631 → 9 → 3 → 2. Nothing in #344's design should be built until we have our own curve on projects we can re-measure.

### Method

Instrument the constant-resolution choke point (`Reflection.resolve_constant_type`) during an ordinary `rigor check` run to collect every *resolved* constant reference, then diff that set against the project's own declared constants (`Scope::DiscoveryIndex`'s `discovered_classes` / `in_source_constants`, restricted to declarations whose source file is inside the analysis root).

This is deliberately the same choke point the tracer-bullet slice would instrument, so the spike doubles as a feasibility check on that injection point.

Report, per corpus target, the candidate count at each successive stage:

1. **Naive** — analysis root = the project's configured `paths:` only
2. **+ widened root** — index `config/` and `lib/` as well (the article's first correction)
3. **+ route roots** — controllers named by `config/routes.rb`, parsed statically by `rigor-rails-routes` (the article's second and third corrections, collapsed: static parsing sees both branches of a conditional route)
4. **+ dynamic-resolution demotion** — candidates whose namespace is touched by `constantize` / `const_get` / interpolated constant names / ERB / YAML, moved to a separate "cannot decide" bucket

Then **sample-adjudicate** the surviving candidates by hand on at least one target: how many are genuinely unused? That ratio, not the raw count, is the number #344's ADR needs.

### Corpus

`~/repo/ruby/rigor-survey/{mastodon,redmine,gitlab,strap}` — all already onboarded with a `.rigor.yml`. **cwd must be the target project, not the rigor repo**, or config discovery and baseline path normalisation both go wrong; borrow the rigor bundle with `BUNDLE_GEMFILE`:

```sh
nix --extra-experimental-features 'nix-command flakes' develop --command bash -c '
  cd ~/repo/ruby/rigor-survey/<project> &&
  BUNDLE_GEMFILE=~/repo/ruby/rigor/Gemfile bundle exec ~/repo/ruby/rigor/exe/rigor <subcommand>'
```

### Harness hazards to respect

[...]

Issue #344, opened 2026-08-13
Reachability report for unused-code candidates (rigor unused)


zonuexe:
## What to build

A whole-project **reachability report** that narrows "which constants in this project are probably unused" down to a list a human can actually check — shipped as its own CLI subcommand (`rigor unused`), **never** as a `rigor check` diagnostic.

Prompted by [Rubydex を用いた未使用コントローラ検出の記事](https://zenn.dev/medley/articles/54e15ed69772ba) (medley, 2026): the author indexed a large Rails app with Shopify's Rubydex, listed controllers with zero resolved constant references, and got **631 candidates**. Three corrections — widening the index beyond `app/` to `config/` and `lib/` (→ 9), reading routes from `Rails.application.routes.routes` instead of regexing `rails routes` text (→ 3), and spotting one conditionally-defined route by hand (→ 2) — left **2** genuinely unused controllers. The author's own conclusion is the shape this feature has to take: a tool for reducing candidates to a checkable number, not a tool that produces answers.

### Why this is a Rigor feature and not "just use Rubydex"

[ADR-21](docs/adr/21-rubydex-evaluation.md) already settled the Rubydex question in three tracks (foundation = reject, backend = defer with triggers, LSP tool = conditional accept). This is a fourth, independent thing: a Rigor-native report built on substrate Rigor already has.

The declaration half is essentially in place:

- `Scope::DiscoveryIndex` (ADR-53) carries project-wide `discovered_classes` / `in_source_constants` / `discovered_class_sources` / `discovered_def_nodes`.
- `Reflection.resolve_constant_type(name, scope:)` performs nesting-aware constant resolution — the exact capability the article credits Rubydex with over `grep` (resolving a bare `Foo` to `Api::V1::Foo`).
- `plugins/rigor-rails-routes` already interprets `config/routes.rb` statically via Prism, with **no Rails runtime** — including `controller:` options, `devise_for`, and `skip_controllers`.

The missing half is a **reverse index**: nothing records which constant reference resolved to which declaration. The nearest existing thing, ADR-46's `Analysis::DependencyRecorder`, records what *type analysis read* across files (`Class#method` granularity, plus negative edges) — not every constant reference, so a constant that is merely mentioned without its type being consumed may produce no edge.

### Design constraints this feature is bound by

[...]

Issue #349, opened 2026-08-13
Plugin-supplied roots, and a controller fact in rigor-rails-routes


zonuexe:
## Parent

#344

## What to build

The plugin-supplied root mechanism, proven end-to-end by its first consumer: `rigor-rails-routes` contributing the controllers named by `config/routes.rb`.

Two halves, landing together so the mechanism is never speculative:

- **Core.** A root-contribution protocol on the plugin surface: a plugin publishes a set of constant names (or declaration sites) that are entry points, carried by the existing `produces:` / `fact_store` fact mechanism (ADR-9) with a `consumes:` ordering so producers run before the reachability sweep. No Rails knowledge enters the core.
- **`rigor-rails-routes`.** The plugin already interprets `config/routes.rb` via Prism with no Rails runtime, and its parser already reads `controller:` options, `devise_for`, and `skip_controllers` — but it publishes only a `HelperTable` of route *helpers*, not the controller constants. Add a controller-set fact and publish it as a root source.

This is the slice where Rigor beats the approach in the source article rather than reproducing it. The article needed `Rails.application.routes.routes` — a booted Rails app — because regexing `rails routes` output dropped routes with trailing metadata; and even then a route defined under a conditional (`get "..." if <cond>`) did not appear in the runtime table and had to be found by reading code. Static Prism interpretation has neither problem: no boot, and **both branches** of a conditional route are visible. Both properties should be pinned by specs, because they are the reason to prefer this design.

## Acceptance criteria

- [ ] A plugin can declare a root source through the documented plugin surface, and the reachability sweep consumes it with the right ordering guarantee
- [ ] `rigor-rails-routes` publishes its controller set as a fact and registers as a root source
- [ ] A controller reachable *only* via a route is not reported as unused
- [ ] Spec: a controller routed under a conditional (`get "..." if <cond>`) is treated as a root — the case the article had to find by hand
- [ ] Spec: no Rails runtime is loaded on the root-supply path
- [ ] Corpus re-measurement on the Rails targets, appended to the spike's note, showing the candidate-count drop attributable to route roots
- [ ] The plugin-surface addition is documented in the manual and the plugin-authoring reference
- [ ] `make verify`, `make check`, `make check-plugins` clean

## Blocked by

- #347 — the tracer-bullet report slice

PR #353, opened 2026-08-15
Add the issue #345 unused-constant probe harness

zonuexe:
Lands the measurement instrument for #345, the spike under #344's reachability-report arc.

Four engine hooks, each behind `UnusedProbe::ACTIVE` — a frozen constant read from `RIGOR_UNUSED_PROBE` at load time, so an unset run is unchanged. `Reflection.resolve_constant_type` is the only place a resolved constant's fully-qualified name is observable (`candidate` at the moment of return, which dies with the frame), hence the small `probe_resolved` indirection; it is private so the ADR-2 read-side facade snapshot in `public_api_drift_spec` is untouched. The other three record the declaration side, the project class-source map and structural edges, and RBS-side type references.

The run-shape hook is the load-bearing one: a non-zero worker count accumulates references in a child process that never reaches the parent, and an attached cache store means a HIT file is never re-analyzed and contributes zero references while its constants still count as declared. Both read as a long list of dead constants. `tool/unused_probe/report.rb` exits 2 rather than printing a number when either holds.

The fixtures are what makes a "no" trustworthy — `gate` covers a bare cross-file reference, a nesting-relative one (`A::B::C` written only as `C` inside `A::B`), an RBS-only reference, and a genuinely unreferenced constant; `cross_file_values` and `superclass_and_defaults` isolate two measured reference-side losses.

## Results so far

On Rigor's own `lib`: 1280 declared constants, 140 candidates, **0 genuine** — five hand-adjudicated spot-checks were all probe artifacts. Full write-up and the corpus survey are on #345.

The finding that matters: **the typing path is not a reference index.** `resolve_constant_type` fires only where the engine needs a constant's type, so a read-but-never-typed constant leaves no trace — cross-file value constants, parameter defaults, lambda-rvalue bodies, superclass positions, and intermediate namespace segments are all invisible. #347 has been corrected accordingly: the reverse index needs a dedicated constant-read-node pass, not a hook on the typing path.

## Note on scope

This is spike scaffolding in engine files, landed deliberately rather than written up and deleted — a docs-only note would throw the instrument away and the next session would rebuild it to re-ask the same question. The hot-path cost is one frozen-constant read per resolved constant. If that is not wanted in `master`, the alternative is to keep this branch unmerged and cite it from the `docs/notes/` write-up.

`make verify` green, `git diff --check` clean.

### DIFF
--- a/lib/rigor/analysis/runner.rb
+++ b/lib/rigor/analysis/runner.rb
@@ -16,6 +16,7 @@ require_relative "../plugin/source_rbs_synthesis_reporter"
 require_relative "../rbs_extended/reporter"
 require_relative "../rbs_extended/conformance_checker"
 require_relative "../reflection"
+require_relative "../unused_probe" # issue #345 spike instrumentation
 require_relative "../type/combinator"
 require_relative "../inference/coverage_scanner"
 require_relative "../inference/parameter_inference_collector"
@@ -229,6 +230,14 @@ module Rigor
         return Result.new(diagnostics: [target_ruby_error]) if target_ruby_error
 
         expansion = expand_paths(paths)
+        # Issue #345 spike instrumentation. Records the run SHAPE the probe's correctness depends on:
+        # a non-zero worker count means per-file references accumulate in a child process and never
+        # reach this one, and an attached cache store means a HIT file is never re-analyzed and
+        # contributes zero references. The report script refuses both. Inert unless the env var is set.
+        if UnusedProbe::ACTIVE
+          UnusedProbe.record_run_shape(files: expansion.fetch(:files), workers: @workers,
+                                       cache: !@cache_store.nil?)
+        end
         @snapshots.reset_for_run
         # Per-run reset of the deferred-discovery memo (see `#ensure_project_discovery`).
         @project_discovery_done = false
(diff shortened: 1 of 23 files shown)


zonuexe:
Follow-up on the hot-path concern raised in the PR body — measured rather than argued.

A/B on `rigor check lib --no-cache --workers=0`, five interleaved rounds, toggling `lib/rigor/reflection.rb` between this commit and `3eb7b4c2` in the same working tree (a worktree baseline cannot resolve its own bundle, so the toggle is the only honest shape):

```
round1  probe=8.00s  pre=7.83s
round2  probe=7.97s  pre=7.97s
round3  probe=7.94s  pre=7.99s
round4  probe=8.06s  pre=7.99s
round5  probe=7.97s  pre=8.09s
```

Means: probe 7.988s, pre 7.974s — **+0.18 %**, against a 0.26 s spread within the `pre` samples alone. The cost is not measurable; YJIT almost certainly inlines `probe_resolved`. The concern was wrong on the performance axis.

What remains is a code-hygiene objection, which is a different and smaller thing: spike scaffolding living in an engine file. Two facts bear on when to act on it:

- **The hook has no future as production code.** #347 established that the typing path is the wrong place for a reference index — `resolve_constant_type` fires only where a type is needed, so the real implementation is a dedicated constant-read-node pass. This hook is purely the spike instrument.
- **It is not finished being used.** #348 and #349 both carry acceptance criteria requiring corpus re-measurement against the #345 baseline. Deleting the instrument now would force those to reconstruct it, which is the failure mode the harness was preserved to avoid.

So: keep it until #349's re-measurement is done, then delete the four hook sites — the note plus this commit preserve the method permanently. Recorded as a cleanup criterion on #344.

If it should come out sooner, the zero-footprint alternative is a `Module#prepend` installed by `unused_probe.rb` only when the env var is set, leaving `reflection.rb` byte-identical to pre-probe. It costs a re-implementation of the candidate loop in the probe file (the resolved `candidate` is a local that dies with the frame, so a wrapper cannot observe it), which can drift from the engine — worth a spec pinning the two to the same answers if we go that way.

Issue #330, opened 2026-08-09
Spec suite leaks Dir.mktmpdir directories; accumulation caused a real ENOSPC


zonuexe:
## Symptom

The spec suite leaks `Dir.mktmpdir` directories. They accumulate under the nix-shell temp dir and are never reclaimed, and on 2026-08-09 they produced a real `Errno::ENOSPC` that failed 33 examples mid-session.

Measured:

- one `make test` (9,093 examples) leaves **~25 entries / ~86 MB**;
- the 405-example `check_rules` covering suite leaves 3 / 21.8 MB;
- a fused mutation census (one suite run per type-survivor) leaves **~1 GB**;
- **1,482 stale directories** had accumulated across `/tmp/nix-shell.*` before the sweep.

Prefixes observed: `rigor-spec-cache-*`, `rigor-spec-workspace-*`, `rigor-spec-sig-*`, `rigor-scanner-spec-*`, `rigor-plugin-spec-cache-*`, `rigor-mutant-*`, plus Ruby-default `d<YYYYMMDD>-*`.

## Why it is worse than housekeeping

The failure mode is not "disk fills up" — it is that the failures **read as unrelated flake**. Under `ENOSPC` the errors surface as `Dir.mktmpdir` / `Dir.chdir` / `File.write` failures scattered across specs that have nothing to do with the change under test, so the natural reading is contention or a bad merge. A mutation census run under those conditions is worse still: a spec that errors out looks like a *non-kill*, so the leak can silently fabricate survivors.

## Root cause and fix direction

The non-block form leaks by construction:

```ruby
dir = Dir.mktmpdir("rigor-spec-cache-")   # never removed
```

The block form removes the directory when the block exits, including on exception:

```ruby
Dir.mktmpdir("rigor-spec-cache-") { |dir| ... }
```

Sweep the suite for non-block `Dir.mktmpdir` calls and convert them, or — where the directory's lifetime is genuinely an example's lifetime rather than a block's — register cleanup in an `after` hook. `let(:tmpdir) { Dir.mktmpdir(...) }` is the common leaking shape and needs the `after` treatment, since there is no block to close.

A belt-and-braces `config.after(:suite)` that removes anything the run created is worth considering, but the per-site fix is the real one — the suite forks and shells out, so a global sweep can race.

## Acceptance

A full `make test` leaves **zero** new entries under the run's temp dir. Assert it: capture the entry count before and after a suite run in CI, or add a spec that fails when the count grows.

## Provenance

[...]

PR #356, opened 2026-08-15
Resolve constants through the ancestor chain, not just lexical nesting

### LINKED ISSUE #354: Constant resolution ignores the ancestor chain, and resolves to a shadowing top-level constant instead
Constant resolution walks only the lexical nesting, never the ancestor chain. Ruby's rule is `Module.nesting` **then the ancestors of the innermost cresting scope** then top level; `Reflection.lexical_constant_candidates` (`lib/rigor/reflection.rb:145`) implements only the first and third parts — it strips `::` segments off the enclosing class path and then falls back to the bare name.

Two consequences, and the second one is a false positive on ordinary Ruby.

## 1. An inherited constant does not resolve

```ruby
class AbstractAdapter
  INHERITED = 42

  def own_scope_control
    INHERITED.no_such_method_own          # fires: undefined method for 42
  end
end

class BazaarAdapter < AbstractAdapter
  def qualified_control
    AbstractAdapter::INHERITED.no_such_method_qualified   # fires: undefined method for 42
  end

  def via_superclass_scope
    INHERITED.no_such_method_inherited    # SILENT — the constant does not resolve
  end
end
```

Both positive controls fire, so this is not a vacuous silence. Real Ruby resolves the third read to `42`.

## 2. When the name is also defined at top level, it resolves to the WRONG constant

```ruby
KEY = "top-level"

class Base
  KEY = 42
end

class Sub < Base
  def read
    KEY.no_such_method_here
  end
end
```

```
lib/all.rb:9:9: error: undefined method `no_such_method_here' for "top-level"
```

Real Ruby: `Sub.new.read # => 42`. Rigor types it as `"top-level"`. The lexical walk's last candidate is the bare name, so the top-level constant wins a lookup that Ruby gives to the ancestor. Every type-dependent diagnostic downstream of that read is then reasoning about the wrong type — this is a false-positive source on correct code, which per AGENTS.md § Implementation Guidelines outranks any worst-case reading.

Reproduce with `rigor check --no-cache --workers=0` on the fixtures above. Note that `rigor type-of` **cannot** reproduce this — it has no project-wide pre-pass — and that a cross-file version of either fixture is confounded by the separate per-file `in_source_constants` limitation (#352), so keep the fixture single-file when isolating this.

## Where it was found

Surfaced by the #345 unused-constant measurement on redmine, where `rescue ScmCommandAborted` inside `class BazaarAdapter < AbstractAdapter` recorded no constant reference at all. It is an artifact class for that report, but the mis-resolution above is an engine defect independent of it.

## Scope note

The fix is an ancestor walk over the innermost cresting scope's superclass chain (and included modules, which Ruby also consults), inserted between the lexical candidates and the bare-name fallback. `discovered_superclasses` and `discovered_includes` are already project-seeded (`runner.rb:1116`), so the data is present. The ordering matters: ancestors must be tried **before** the bare name, or case 2 stays wrong.

## Acceptance criteria

- [ ] Bare reference to a superclass-defined constant from a subclass body resolves, with the same-scope and fully-qualified controls kept in the spec so silence is never read as evidence
- [ ] The shadowing case resolves to the ancestor's constant, not the top-level one
- [ ] Included-module constants resolve too (Ruby consults the full ancestor chain, not just superclasses)
- [ ] A constant that exists in neither the lexical nesting nor the ancestor chain still falls back to the bare name unchanged
- [ ] Corpus FP diff across the survey targets: no new firings
- [ ] Behaviour recorded in the relevant `docs/type-specification/` document in the same commit
- [ ] `make verify`, `make check`, `make check-plugins` clean

zonuexe:
Closes #354.

Ruby looks a constant up in three steps: each entry of `Module.nesting`, then **the ancestors of the innermost cresting scope**, then the top level. `Reflection.resolve_constant_type` implemented the first and third and skipped the second.

The missing step is not merely a lost resolution. Because the bare-name fallback still ran, a name that exists **both** on an ancestor and at top level was answered by the top-level constant — the one Ruby does not choose:

```ruby
KEY = "top-level"
class Base
  KEY = 42
end
class Sub < Base
  def read = KEY      # Ruby: 42.  Rigor before this change: "top-level"
end
```

Everything downstream of that read was then reasoning about the wrong type — a false positive on correct code.

## Scope

Breadth-first, mixins ahead of the superclass at each level (Ruby's ancestor order), terminating on a cyclic graph. **Only project ancestors participate**: `superclass_of` / `includes_of` carry as-written names, each is resolved against the subclass's lexical nesting, and a name resolving to no discovered project namespace is dropped. `class Model < ActiveRecord::Base` therefore contributes nothing, and a constant owned by a gem superclass is still reachable only by its full name. Widening to the RBS ancestor graph is a separate question with its own FP surface; the boundary has a spec pinning it.

Step 2 runs whenever the lexical candidates all miss — the common case for a core-class reference (`String` inside `class Foo`) — so the ancestor list is memoised per run on the ADR-84 WD2 generation token. A run with the ADR-46 dependency recorder active **bypasses** the memo: `superclass_of` / `includes_of` record a cross-file edge per consumer file, and a run-scoped memo hit would skip the recording and under-record the edge for every later file.

## Corpus gate

Zero new firings across eleven survey targets — redmine, mastodon, conference-app, mangrove, kramdown, liquid, faraday, oj, ox, haml, rubocop-ast — with diagnostic sets byte-identical in both directions.

Eleven byte-identical results are also exactly what a broken A/B looks like, so the toggle mechanism was separately shown to flip the fixture from `"top-level"` to `42`. The zero is real, not a harness that never swapped the file.

## The fix is not inert on real code

Measured with the #345 probe: redmine gains **164 resolutions** and retires exactly one candidate from the #345 baseline — `Redmine::Scm::Adapters::AbstractAdapter::ScmCommandAborted`, which is precisely the single artifact that adjudication had attributed to this gap — with **no new candidate appearing**. Mastodon gains six resolutions.

## Verification

`make verify` exit 0 (includes `make check` and `make check-plugins`), `make docs-check` exit 0, `git diff --check` clean. Behaviour recorded in `docs/internal-spec/inference-engine.md` § Lexical constant lookup in the same commit; `[Unreleased]` changelog entry written user-facing.


zonuexe:
CI was red on `Tests (Ruby 4.0)`, and it was not this change.

`0/9153 examples failed` — every test passed. The run failed in `after(:suite)`, on issue #330's temp-residue guard:

```
spec suite leaked 1 temp entry into /tmp/rigor-spec-run-20260815-2933-umctd4:
mutant20260815-2933-rrp2ct.rb
```

Source: `spec/rigor/protection/closure_kill_oracle_spec.rb:184` used `Tempfile.new(["mutant", ".rb"])`, which reclaims on finalization — so whether the file is still in the private root when the residue check runs is a GC-timing coin flip. Introduced in a656fe05 (2026-08-01), unrelated to constant lookup. The rest of the suite already uses `Tempfile.create` with a block; this was the one holdout. Fixed in a separate commit so the history stays honest about the two concerns.

**On the strength of that identification** — the leaked name matches `Tempfile.new(["mutant", ".rb"])`'s `mutant<date>-<pid>-<random>.rb` shape exactly, and it is the only such call in the suite. But I could **not** reproduce the leak locally: running that spec file alone passes with the old form too, because five examples are not enough to make GC timing bite. So the fix rests on the mechanism, not on a before/after observation, and my positive control for it was inconclusive. If CI goes red again on the same guard, that assumption is what to revisit first.

Issue #533, opened 2026-08-31
Minor verified dispatch gaps from the 2026-09-01 corpus sweep (eight small mechanisms)


zonuexe:
Grab-bag of small, individually-verified dispatch gaps from the 2026-09-01 corpus opacity sweep (branch `opacity-sweep-20260901` carries the repros; per-target reports cite exact sites). Each is one bounded fix; none justifies its own issue:

1. **Heredoc-folded Constant[String] dispatch inconsistency** — a multi-line-heredoc-folded `Constant[String]` fails `#squish` dispatch while the identical type from a plain literal resolves; `.upcase` on the same receiver folds (mastodon, 28 sites; `repros/mastodon_controls3.rb`).
2. **`alias_method :new_name, :old` sends unmodeled** — the `alias` KEYWORD works, the method form does not (liquid i18n `#t`, 33 sites; `repros/typeof_probe7.rb`).
3. **`send(:literal_symbol, ...)`** never resolves though the selector is statically known at virtually every site (protobuf, 84 sites — the private-boundary idiom).
4. **Refined-receiver lookup does not erase to the base class** — `RUBY_VERSION != "1.0"` answers Dynamic on a `non-empty-string` receiver (parser 13, net-ssh 15 sites); erase `Refined` to its base for method lookup.
5. **`Proc#[]` call form** unhandled on literal procs (net-ssh, 6 sites).
6. **rbs-core toplevel `::Queue` / `::SizedQueue` aliases missing** — bundled core RBS declares only `::Thread::Queue`; plain `Queue.new` answers Dynamic (~55 sites across the algorithm corpora).
7. **Conditional-superclass / rescue-LoadError implementation selection** — `Impl = case RUBY_ENGINE ... end` superclasses (concurrent-ruby `synchronize` 186 sites, AtomicFixnum 43) and `Containers::Deque = CDeque rescue RubyDeque` aliases (algorithms 30): fold the statically-decidable engine predicate, or join only the resolvable arm instead of dragging the join to Dynamic.
8. **Empty-array index mutation keeps the empty-tuple shape** — `stack = []; stack[top] = cs` leaves `[]` so later index reads answer Dynamic (mail's generated ragel tables, 117 sites); index-write widening should promote the empty tuple to `Array[T]`.

**Gate per item:** `make verify` + self-check gates; corpus FP diff; paired must-fire/must-still-succeed controls (the formatter-splat lesson).

zonuexe:
[...]

PR #550, opened 2026-08-31
Erase refinements to their base for RBS lookup; widen is_a? guards on unknown classes

zonuexe:
Part of #533 (item 4 — refined-receiver lookup), plus the narrowing companion its own corpus gate demanded.

**Refined/Difference lookup erasure.** `non-empty-string` is `Difference[String, ""]`; neither `Refined` nor `Difference` had a `receiver_descriptor` arm in `RbsDispatch`, so any method the refinement catalog does not promote declined the whole dispatch to Dynamic — `RUBY_VERSION != "1.0"` / `RUBY_PLATFORM == "java"` read `Dynamic[top]` (parser 13 sites, net-ssh 15). Lookup now erases to the base carrier; the catalog tier's refinement promotions (`upcase` → `uppercase-string`) run above this tier and still win (control spec).

**`is_a?` narrowing on an unorderable class.** The erasure made one new firing reachable and its root is a standalone narrowing bug: `narrow_nominal_to_class` KEPT the old bound on `:unknown` ordering (guard class = an RBS-less gem class). Keeping the bound contradicts the guard — concurrent-ruby's `break done.value if done.is_a?(Concurrent::Maybe)` kept `Array` through the guard and fired `undefined-method` on `.value`. The `:unknown` arm now answers Dynamic (the guard proved membership in a class the engine cannot name; the old knowledge is destroyed). `:subclass` still preserves the bound; `:superclass`/`:disjoint` unchanged; the old behavior's pinning spec is rewritten to pin the new contract with the rationale.

Corpus diff over the 11 gate targets: **0 added / 0 removed** (the interim single firing was neutralized by the narrowing half). `make verify` / `git diff --check` green.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,8 @@ Older release notes are archived under [`docs/`](docs/) when the leading version
 
 ### Fixed
 
+- **[engine]** Methods on refined strings (`RUBY_VERSION != "1.0"`, `RUBY_VERSION.split(".")`) now resolve through the underlying `String` instead of losing the whole call to untyped, and an `is_a?` guard against a class Rigor cannot see no longer keeps the old type on the guarded branch — the guard's own proof wins ([#550](https://github.com/rigortype/rigor/pull/550), [#533](https://github.com/rigortype/rigor/issues/533)).
+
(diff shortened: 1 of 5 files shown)


zonuexe:
Bonus corpus effect verified with the sweep harness's `mastodon_controls3.rb`: the multi-line-heredoc `#squish` inconsistency (#533 item 1, 28 mastodon sites) is the same mechanism — the heredoc folds to the `literal-string` Refined/Difference carrier, and its dispatch declined for any method the catalog does not promote. On this branch `<<~CSS.squish` resolves to String through the plugin's RBS; on master it reads Dynamic with an unsupported_syntax origin.

Issue #569, opened 2026-09-01
rigor-activerecord disables all model typing when db/schema.rb is absent instead of degrading to a columns-less index


zonuexe:
Found by the 2026-09-01 post-campaign opacity re-run (the sweep probe re-run against the fully merged master, `a7e5d805`): the single largest attributable opacity cluster on redmine is not an engine or issue-tracked gap at all — **`rigor-activerecord` disables 100% of its model typing whenever the target has no committed `db/schema.rb` / `db/structure.sql`**, emitting only a load-error warning.

**Mechanism.** The plugin's private `model_index` accessor (`plugins/rigor-activerecord/lib/rigor/plugin/activerecord.rb`, ~line 512) is an all-or-nothing gate: `schema_table_or_nil` returning nil short-circuits the entire index. But the schema is only needed for `columns_for(table_name)` — `ModelIndex::Entry#table_name` (`sti_table_name` / `Inflector.tableize`) and everything `ModelDiscoverer#discover` produces (scopes, associations) are schema-independent, and `finder_return_type` / `class_scope_return_type` would work off those alone. Redmine gitignores `db/schema.rb` (`/db/schema.rb` in its `.gitignore` — the DB-agnostic Rails pattern that ships raw migrations only), so every modeled AR call answers Dynamic there.

**Evidence** (verified on the real CLI, current master): `rigor check app/models/issue.rb` on redmine prints `warning: rigor-activerecord: schema file 'db/schema.rb' (or 'db/structure.sql') not found; AR call checks skipped [plugin.activerecord.load-error]`; even `singleton(Project)#find` — a call the plugin fully models — is opaque on redmine (15 sites). On mastodon, which commits `db/schema.rb`, the plugin loads and **neither `table_name` nor `.where`/`.visible` appears anywhere in the top-80 opaque pairs**. Probe counts on redmine: `table_name` family **517** named-receiver-opaque sites (`singleton(Issue)#table_name` 137, `singleton(Project)#table_name` 98, …) plus **~127** first-hop `.where`/`.visible`/`.find` sites that `class_call_return_type`/`class_scope_return_type` already model.

[...]

PR #571, opened 2026-09-01
Fold JSON.generate and JSON.pretty_generate to String

### LINKED ISSUE #570: Fold JSON.generate / JSON.pretty_generate to String — upstream RBS declares untyped
From the 2026-09-01 post-campaign opacity re-run: `singleton(JSON)#pretty_generate` is 25 named-receiver-opaque sites on rigor-lib and 3 on herb, and the cause is upstream — `references/rbs/stdlib/json/0/json.rbs:1170` declares `def self?.pretty_generate: (_ToJson obj, ?options opts) -> untyped` (same for `generate`). Verified the signature IS loaded and translated faithfully (same-file control `JSON.pretty_generate({})` answers `Dynamic[top], erased: untyped` — not a library-loading or overload-selection decline), so this is honest RBS translation, closable only by a Rigor-side fold.

**Fix direction:** add a `JsonFolding` singleton tier in `lib/rigor/inference/method_dispatcher/` modeled on `CgiFolding` / `ShellwordsFolding`, returning `Nominal[String]` for `JSON.generate` / `JSON.pretty_generate` (both always return String or raise). Monotone over today's `untyped`; the constant-fold tier for literal arguments can stay out of scope.

**Gate:** `make verify` + self-check gates; must-fire control (`JSON.pretty_generate({})` types String) + must-still-succeed control for an unrelated JSON method (`JSON.parse` stays untyped); corpus FP diff expected byte-identical (a return-type narrowing from untyped cannot add firings on its own, but run it anyway per the standing rule).

zonuexe:
Implements #570, from the 2026-09-01 post-campaign opacity re-run: `JSON.pretty_generate` is declared `-> untyped` in upstream RBS (`stdlib/json/0/json.rbs`), so all 25 rigor-lib + 3 herb call sites widened to `Dynamic[top]`. This adds a `JSONFolding` singleton tier (modeled on `CGIFolding`/`ShellwordsFolding`, registered in the `STDLIB_SINGLETON_FOLDERS` table) returning `Nominal[String]` for `generate`/`pretty_generate` with 1–2 args — monotone over today's answers. Return-TYPE fold only: the tier deliberately does not evaluate the argument (a literal-argument constant fold can layer on later without changing it).

Note for review: upstream RBS already declares `generate -> String`, so the `generate` half is redundant defense-in-depth today; `pretty_generate` is the actual gap. Both are folded for symmetry per the issue.

Specs: 12 examples mirroring `cgi_folding_spec.rb` — must-fire for both methods and the 2-arg form, paired declines for `JSON.parse`/`JSON.dump`, wrong arity, non-Singleton and instance receivers, wrong singleton class, and Dynamic receiver.

Gates: full `make verify` + 11-target corpus FP diff run serially at landing (expected byte-identical — the fold is monotone); results will be posted below before merge.

Closes #570.

### DIFF
--- a/changelog.d/added/json-generate-fold.md
+++ b/changelog.d/added/json-generate-fold.md
@@ -1,1 +1,1 @@
-- **[engine]** `JSON.generate` and `JSON.pretty_generate` now type as `String` instead of widening to `untyped`, closing a named-receiver-opaque gap upstream RBS left on `pretty_generate`. ([PR #TBD](https://github.com/rigortype/rigor/pull/TBD))
+- **[engine]** `JSON.generate` and `JSON.pretty_generate` now type as `String` instead of widening to `untyped`, closing a named-receiver-opaque gap upstream RBS left on `pretty_generate`. ([#571](https://github.com/rigortype/rigor/pull/571))
--- a/changelog.d/added/json-generate-fold.md
+++ b/changelog.d/added/json-generate-fold.md
@@ -1,0 +1,1 @@
+- **[engine]** `JSON.generate` and `JSON.pretty_generate` now type as `String` instead of widening to `untyped`, closing a named-receiver-opaque gap upstream RBS left on `pretty_generate`. ([PR #TBD](https://github.com/rigortype/rigor/pull/TBD))
--- a/lib/rigor/inference/method_dispatcher.rb
+++ b/lib/rigor/inference/method_dispatcher.rb
@@ -749,7 +749,7 @@ module Rigor
       # receiver and their relative trial order was never observable. Compiling them into a
       # class-name table turns nine no-op trials per call into one Hash read, skipped entirely
       # when the receiver is not a `Singleton` (the overwhelmingly common case). The table sits
-      # where the nine sat in the old flat list: after ShapeDispatch, before KernelDispatch.
+      # where the original eight sat in the old flat list: after ShapeDispatch, before KernelDispatch.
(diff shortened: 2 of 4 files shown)


zonuexe:
Landing gates, all green:

- **`make verify`** (worktree, full suite): exit 0 — 10,219 examples across the sharded suite, lint, `make check`, `make check-plugins` all clean.
- **11-target corpus FP diff** (base = master `0aa75697`, `--no-baseline --no-cache`, same config both arms): byte-identical on slim / tdiary-core / haml / kramdown / liquid / net-ssh / concurrent-ruby / textbringer / mail / mastodon. redmine shows +3/−7 on the `plugin.activerecord.load-error` warning only — the schema-missing disclosure (#569's subject) anchors to whichever file first triggers plugin load, which jitters run-to-run under parallel workers; zero real diagnostic changes.
- **Critical review** (independent reviewer): APPROVE, no blocking findings. Two nits fixed in `d9d41ccf` (a historically false "nine sat in the old flat list" comment, a sibling-template doc heading); one report-only note carried: the fold shadows a hypothetical project-defined top-level `JSON` singleton, the same precedent every landed stdlib singleton folder carries (verified `CGI` behaves identically; namespaced `MyApp::JSON` is safe — resolution produces fully-qualified singletons and the fold declines).
- CI: all checks pass on `d9d41ccf` (the earlier shard-2 failure was the GC-dependent `rigor-mutant-*` tempfile flake, reproduced-then-passed on identical code; filed as #572).

Issue #575, opened 2026-09-01
def.ivar-write-mismatch fires on per-action controller ivars — divergent writes in mutually exclusive Rails actions are never live together


zonuexe:
Surfaced by the #569 landing gate on redmine (4 new `def.ivar-write-mismatch` warnings in roles/workflows controllers, e.g. `@roles` assigned `Role.sorted.to_a` (Array) in `index` and `Role.sorted` (Relation) in `permissions`). The writes genuinely diverge, so the rule fires per its contract — but on a Rails controller the contract's premise is a category error: each request instantiates a fresh controller and runs exactly ONE action, so per-action `@ivar` writes in different actions are never live together and the divergence has no runtime consequence. This is the same shape as the `db/migrate/` carve-out rigor-activerecord already ships (validating evolving state against a single snapshot).

Not caused by the #569 change — the rule pre-exists and only became reachable once the Relation types got concrete (both arms were `Dynamic` before, which the rule's envelope skips). Expect the same class of warning to grow as any precision lever lands on Rails apps.

**Proposed direction (rule-level, cross-plugin):** decline the mismatch when the conflicting writes sit in different, mutually exclusive ACTION methods of a controller — `rigor-actionpack` already knows what a controller is and what its actions are, so the natural shape is an ADR-9 fact (or an engine hook the plugin feeds) marking per-action write scopes. A plugin-side suppression that degrades the Relation type instead would be the wrong trade. Needs the usual paired controls: the carve-out must not silence a genuine same-action or shared-helper divergence.

Until then these land in the acknowledge-mode baseline on survey targets (4 sites on redmine).

Issue #573, opened 2026-09-01
rigor-shoulda-matchers reads the :model_index fact as a ModelIndex object, but the fact publishes a Hash


zonuexe:
Found while auditing the ADR-9 `:model_index` fact's consumers for PR-#569 work (the reduced-mode index deliberately withholds the fact, which forced reading every consumer): `rigor-shoulda-matchers` consumes the fact as if it were a `ModelIndex` OBJECT, but the fact publishes a plain Hash.

The consumer calls `index.find(anchor)` and then `entry.column?(…)` on the result. On the published Hash, `Hash#find` with one non-block argument… does not do a keyed lookup at all (with no block it returns an Enumerator; the `ModelIndex#find` object method is what the code shape expects), so the real cross-plugin path can never produce the entry it thinks it is checking. The plugin's own spec stubs a `ModelIndex`-shaped object for the fact and therefore never exercises the published-Hash shape — the same green-gate-that-cannot-see trap as the doc-extractor postmortem.

Impact: `rigor-shoulda-matchers`' `unknown-column` checks over `have_db_column` / `validate_presence_of` are likely inert (or worse, mis-resolving) on every real project, silently. Fix direction: publish or consume one shape consistently — either the fact carries the `ModelIndex` object (check Ractor/Marshal constraints on ADR-9 facts) or the consumer indexes into the Hash by key; then replace the spec's stub with the actual fact payload the producer publishes, so the contract is pinned end-to-end.

Gate: a cross-plugin integration spec (activerecord publishes → shoulda-matchers consumes) with a must-fire unknown-column case and a must-not-fire known-column sibling; `make check-plugins` clean.

Issue #572, opened 2026-09-01
ClosureKillOracle mutant Tempfile lifetime is GC-dependent — the #330 residue guard flakes on CI


zonuexe:
Hit on PR #571's first CI run (Tests shard 2/3, run 33509942004): the issue-#330 residue guard failed with `spec suite leaked 1 temp entry into /tmp/rigor-spec-run-…: rigor-mutant-….rb`; the rerun of the identical head passed, and two local full-suite runs of the same tree never tripped it — a timing flake, not a regression.

The mechanism is the one `spec/rigor/protection/closure_kill_oracle_spec.rb` already documents (~line 104): `ClosureKillOracle`'s mutant `Tempfile` is reclaimed only by its finalizer, i.e. GC-timing-dependent, and the spec's mitigation registers every oracle it builds for release after the example. The CI failure shows one creation path still escapes that registration (or the release path itself doesn't unlink early enough under shard-2 load) — the guard saw exactly one leaked `rigor-mutant-*.rb`.

Fix direction: make the oracle's tempfile lifetime deterministic instead of GC-reliant — an explicit `#release`/`close!` in the oracle's own teardown path (or `Tempfile.create` with a block-scoped file), so no spec needs a side-table of built instances at all. The spec's `built_oracles` mitigation can then shrink to an assertion.

Gate: the #330 residue guard is the spec; a stress rerun of the protection spec directory should show zero `rigor-mutant-*` residue.

PR #576, opened 2026-09-01
Degrade rigor-activerecord to a columns-less index when the schema is absent

### LINKED ISSUE #569: rigor-activerecord disables all model typing when db/schema.rb is absent instead of degrading to a columns-less index

zonuexe:
Implements #569: `rigor-activerecord`'s `model_index` was all-or-nothing on `db/schema.rb` / `db/structure.sql`, so a schema-less checkout (redmine gitignores `db/schema.rb` — the DB-agnostic Rails layout) silently got ZERO model typing. The plugin now degrades to a **columns-less `ModelIndex`**: table names, scopes, associations, enums and finder returns stay active off `ModelDiscoverer` output alone; every column-dependent surface declines exactly as an unknown column does today, and the ADR-9 `:model_index` fact is withheld unless columns are known (closing the `rigor-actionpack` `unknown-permit-key` / `rigor-shoulda-matchers` `unknown-column` blast radius — auditing that consumer surface also found #573). The missing-schema disclosure drops `warning` → `info` (reduced mode is supported input, not a misconfiguration; unreadable/unparseable schema stays `warning`).

Second commit adds `table_name` / `quoted_table_name` to the recognised class-side surface. Deviation from the issue text, deliberate: the `Constant[name]` pin is gated three ways (declared in source / schema-corroborated / not computed-by-the-class) because `table_name_prefix`/custom inflections make a bare-inflection `Constant` a wrong type acted on downstream; uncorroborated names answer `Nominal[String]` — monotone, still closes the opacity. `quoted_table_name` is always `String` (adapter-dependent quoting).

Third commit fixes an engine FP the redmine gate surfaced: `call.wrong-arity` / `call.argument-type-mismatch` never consulted ADR-26's open-receiver surface, so `Issue.visible.open` — `open` is a DECLARED scope on Issue, chained on the now-typed Relation — resolved through the ancestry to `Kernel#open` and errored on correct code (3 sites). Both rules now decline on an unbounded receiver surface **only when the signature is inherited** (`defined_in` ≠ the open class): `Relation#limit`, declared in the plugin's own RBS, still fires. Recorded as ADR-26 WD7; `docs/internal-spec/plugin.md`'s `open_receivers` row updated in the same commit.

**Corpus gate** (base = post-#571 master, full 11-target set, `--no-baseline --no-cache`): 10 targets byte-identical (incl. mastodon — schema-present projects untouched by construction). redmine: **+215 `model-call` :info** recognition traces (the plugin lives), **−3 pre-existing `call.undefined-method` FPs removed**, +8 `call.possible-nil-receiver` (worst-case-sound association nilability, acknowledge-mode baseline lane), +4 `def.ivar-write-mismatch` on per-action controller ivars — adjudicated as a rule-level category error and filed as #575, not suppressed here. Zero errors on correct code after the WD7 fix.

Specs: 123 plugin-integration examples incl. reduced-mode must-fire siblings against the schema-present arm, three-way `table_name` pin gating with discrimination verified by deliberate guard removal, and the WD7 pair (must-not-fire chained-scope `open`, must-still-fire `Relation#limit` + a same-run `Array#rotate` control). Full `make verify` green on the rebased tree.

Closes #569.


zonuexe:
Landing gates, all green:

- **`make verify`**: exit 0 on the final head (run twice — after the WD7 engine fix and after the review-round commits).
- **Corpus FP diff** (base = post-#571 master, full 11-target set, `--no-baseline --no-cache`): 10 targets byte-identical incl. mastodon (schema-present untouched); redmine +215 `model-call` :info recognition traces, **−3 pre-existing `call.undefined-method` FPs removed**, +8 worst-case-sound `call.possible-nil-receiver` (association nilability, baseline lane), +4 `def.ivar-write-mismatch` adjudicated as a rule-level category error and filed as #575. The 3 `call.wrong-arity` errors the first arm surfaced (`Issue.visible.open` → `Kernel#open`) are fixed by the WD7 commit; the declared-only pin round left redmine byte-identical to the approved composition.
- **Critical review** (independent): first pass REQUEST-CHANGES with two verified blockers — (1) the inflection+schema-corroboration `Constant` pin mis-pinned under `table_name_prefix` + name collision (now: declared-literal-only pin, lane deleted), (2) ADR-26 WD7 overclaimed zero loss (`relation.each_slice` was catchable and is now silent — the ADR now records the real trade). Re-review on the fix commits: **APPROVE**, all findings verified resolved against the reviewer's own probes.
- Follow-ups filed from this landing's audit trail: #573 (shoulda-matchers fact-shape), #575 (controller-ivar rule carve-out), #577 (absence-edge cache invalidation).
- CI: green (one #572 tempfile-flake rerun; a deterministic-lifetime fix is in progress on a separate branch).

Issue #574, opened 2026-09-01
possible-nil-receiver's witness gate counts a named-but-RBS-less arm as answerable, FP-blocking nilable plugin nominals (Parameters#[])


zonuexe:
Split out of #534 item 1's adjudication (branch `actionpack-parameters-index-read`): typing `ActionController::Parameters#[]` as `Parameters | nil` — the honest contract — is currently FP-blocked by a rules-level vacuity, and fixing THAT is the only path that reopens the ~1,077-site `params[:k]` mass (the corpus's single largest named-receiver-opaque pair).

**Mechanism.** `union_has_nameable_non_nil_arm?` (`lib/rigor/analysis/check_rules.rb:1358`) is the witness gate for `call.possible-nil-receiver`: the union must carry a nameable non-nil arm for the rule to fire. But `method_present_anywhere?` (`:1380`) answers "present" for a **named-but-RBS-less** arm only because nothing is known about it — so a plugin-declared lenient nominal like `Parameters` counts as a full witness, and `q = params[:q]; q.strip` fires at error severity on correct code (measured 3/3 on the fixture; the four guarded forms narrow correctly and stay silent). That is arguably the same vacuity the gate's own docstring says it exists to prevent: "we do not know this type's surface" is being read as "the surface is present".

**Proposed direction.** Treat a named arm with no RBS and no discovered defs as *unknown* rather than *answerable* in the witness gate — the nil-arm then has no confirmed witness and the rule declines, which is exactly the FP-first ordering. Cost to measure before landing: the documented `Dynamic | String | nil` firing (the gate's current must-fire example) would stop firing, and any corpus site whose witness is an RBS-less gem class goes quiet — needs its own corpus FP/FN diff, not a drive-by (which is why it was not folded into the #534 branch).

**Payoff once landed:** `#[] -> Parameters | nil` becomes typable FP-free (the chain carries: `params[:user][:name]` → `Parameters?`, dispatcher strips the nil arm for the receiver gate), reopening #534 item 1's remainder.

Evidence: the #534 item-1 branch's spec file carries both rejected arms as paired red/green controls (Arm A `-> Parameters` fails exactly the truthiness-fold pin, Arm B `-> Parameters | nil` fails exactly the possible-nil pin, each with a must-fire sibling proving the fixture can raise both rules).

zonuexe:
[...]

Issue #534, opened 2026-08-31
Rails-surface plugin gaps measured on redmine/mastodon: Parameters#[]/expect/slice, Rails readers, Duration, perform_async, concern scopes, AMS object


zonuexe:
> **Status (2026-09-16).** Only items 5 and 6 are still open. The table below is the live list; the original sweep text follows unchanged for the measurements. An agent that picks this issue up should take item 5 or item 6 and nothing else.
>
> | item | status | where |
> | --- | --- | --- |
> | 1 `Parameters#expect` / `#slice` | landed | #548 |
> | 1 `Parameters#[]` | landed (nilable) | #771 (needed #574) |
> | 1 `FlashHash#now`, `keep` / `discard`, request predicates | landed | #592 |
> | 1 `FlashHash#[]=` / `Session#[]=` | no rule needed — assignment semantics already type them | #592 |
> | 1 `FlashHash#[]` / `Session#[]` reads | both arms rejected, waits on a human ruling: no RBS-less nominal can name the value without re-raising `possible-nil-receiver` on `session[:user_id]` | #592 |
> | 1 `Request#format` | typed in #592, withdrawn, split out | #594 (`ready-for-human`) |
> | 2 Rails singleton readers | landed | #585 |
> | 3 Duration multipliers | landed as a `dynamic_return`; the `-> untyped` in `core_ext.rbs` is deliberate, the RBS question is #660 | #585 |
> | 4 sidekiq `perform_async` / `perform_in` / `perform_at` → `String` | landed | #585 |
> | 5 concern-declared scopes | **open** — `lookup_scopes` still reads only the model class body; the `included do` descent exists for `serialize` / `mount_uploader` overrides only | — |
> | 6 AMS `object` recognizer | **open** — no plugin owns ActiveModelSerializers | — |
> | 7 framework constants | landed | #974 |

---

The plugin-territory (E) yield table from the 2026-09-01 corpus opacity sweep, measured on redmine + mastodon (branch `opacity-sweep-20260901`, reports `redmine.md` / `mastodon.md`). Site counts are named-receiver pairs from the plugin-aware probe; the AR scope-chain story is #460 and not repeated here.

[...]

PR #578, opened 2026-09-01
Type the never-nil ActionController::Parameters chain links

zonuexe:
Implements the `ActionController::Parameters` half of #534 item 1, with the `#[]` question adjudicated at the rules level rather than assumed.

**What lands:** `STRONG_PARAMS_CHAIN_METHODS` gains the 13 methods whose Rails implementation returns a `Parameters` on every path — `except`, `without`, `extract!`, `slice!`, `merge`, `merge!`, `reverse_merge`, `reverse_merge!`, `with_defaults`, `with_defaults!`, `compact`, `compact_blank`, `deep_dup` — each contract read off rails v8.1.0.beta1 source, not recalled. Deliberately excluded, with reasons recorded in the plugin comment: `compact!` (nil when unchanged), `select`/`reject`/`filter`/`transform_*` (Enumerator without a block), `select!`/`keep_if` (doc contradicts body), `dig`/`fetch`/`to_h`.

**What does NOT land — `#[]` stays Dynamic, by measurement.** Both candidate typings were run against a fixture controller: `-> Parameters` makes `mode = params[:full] ? :full : :short` fold and draws `flow.always-truthy-condition` on the live guard (a wrong type on correct code); `-> Parameters | nil` carries the chain and fixes the folds, but the assigned form `q = params[:q]; q.strip` fires `call.possible-nil-receiver` at error severity (3/3 in the probe) while all four guarded forms narrow correctly — the witness gate counts the named-but-RBS-less `Parameters` arm as answerable. Both invert the FP-first ordering, so neither ships. The rules-level change that would reopen `#[]` FP-free is filed as #574; the spec file pins both rejected arms as paired red/green controls so the adjudication is executable, not prose.

Leniency-gate context (why the 13 links are FP-safe): `call.undefined-method`/arity/argument rules decline on `ActionController::Parameters` because Rigor ships no RBS for it (`check_rules.rb`'s `rbs_class_known?` gates), so re-typing values that are already Parameters is monotone-additive; fixture proves core `Hash#merge`/`Array#slice`/`Kernel#require` keep their answers.

**Gates:** full `make verify` exit 0 on the rebased tree; corpus FP diff (base = post-#576 master, `--no-baseline --no-cache`) — redmine and mastodon both **byte-identical** (+0/−0). Specs: 7 added in the actionpack integration spec (13 links + 3 chains must-fire; excluded-methods and core-collision must-not-fire; the `#[]` adjudication controls verified to go red under the typing each rejects).

Refs #534.


zonuexe:
Landing gates, all green:

- **`make verify`**: exit 0 on the rebased tree (post-#576 master). The review-nit round added only comments and one must-fire spec sibling; the touched spec file re-run green (54 examples) and CI covers the full suite on the final head.
- **Corpus FP diff** (base = post-#576 master, `--no-baseline --no-cache`): redmine and mastodon both **byte-identical** — the 13 chain links are receiver-gated and monotone-additive, so the typing gain shows in precision, not diagnostics.
- **Critical review** (independent): APPROVE — all 13 Rails contracts re-verified against v8.1.0.beta1 source line by line (plus cross-major drift v6.0→v7.1: no contract changes on methods that exist there), both rejected `#[]` typings re-executed via throwaway arm-plugins (Arm A draws the truthiness fold, Arm B draws error-severity possible-nil — corroborating #574), the core-collision claim re-proven with the reviewer's own fixture. Three nits (stale line citation, a missing must-fire sibling for the fold rule, and `reject!`/`delete_if`/`compact_blank!` being admissible-but-unlisted) addressed in `4a584096`.
- CI: green (one #572 tempfile-flake rerun; the deterministic-lifetime fix lands next).

Issue #580, opened 2026-09-01
Mutation widening is a one-way door: a widened Nominal carrier never re-joins later stores' value types


zonuexe:
Split out of the #560 implementation (branch `array-added-value-join`): mutation widening is a ONE-WAY door — the first content mutation replaces the literal carrier (`Tuple`/`HashShape`) with a `Nominal`, and `widen_for_mutator`'s `case type` has no `Nominal` arm, so every LATER store into the same collection is invisible to content evidence. A step-table repro: `h = {}; h['a'] = {}; h['b'] = "s"; h['c'] = []` ends as `Hash[String, Hash[…]]` — the String and Array stores never joined.

The #560 change contains the FP consequence (Hash value/key parameters stay gradual after widening; Array keeps a precise per-position join), but the Array side still drops later arms: `a = []; a.push(1); a.push("s")` → `Array[Integer]` — an incomplete union that is merely imprecise today (positions accumulate; no corpus FP found), yet it is exactly the stale-evidence family (#540/#541/#544/#560) one precision lever away from firing.

Fix direction: make a MUTATION-WIDENED `Nominal` re-joinable. That needs a provenance mark distinguishing widened-from-literal nominals from RBS/declaration-sourced ones — re-joining the latter is the #561 blast radius (haml's declared `Array[:multi]`-family sigs drew 8 return-type-mismatch FPs under an unscoped widen). With the mark, later mutator args join into the widened carrier's element union on both carriers, and the Hash gradualization in #560's fix can be revisited (a complete union restores per-key honesty).

Gate: the #560 branch's step-table fixture extended to assert later arms present; the ADR-56 `loop_body_fixpoint` golden snapshot must keep `Array[Integer]`; corpus FP diff; haml's temple fixture must stay at 0.

zonuexe:
Two corrections from the PR #581 review (independent reviewer, probes preserved in that review's notes):

[...]

PR #581, opened 2026-09-01
Join a straight-line mutation's added value into content evidence

### LINKED ISSUE #560: Index compound write keeps stale element evidence: t[0] += 5 leaves Array[1 | 2] and fires always-falsey on correct code
Found by probing #532's residual (b) on 2026-09-01. **Four lines of correct code draw a false `always-falsey`, on current master** (pre-dates the sweep's PR queue; also reproduces on the all-in integration):

```ruby
t = [1, 2]
t[0] += 5
c = t[0]
puts "six" if c == 6   # flow.always-truthy-condition: always falsey — but the runtime value IS 6
```

**Mechanism.** The index compound write (`Prism::IndexOperatorWriteNode`) retracts the tuple's arity (#504's mutation widening) but keeps the ORIGINAL element evidence: `t` continues as `Array[1 | 2]`, so `t[0]` reads `1 | 2` while the runtime holds `6`. The written value's type never joins the element set — `StatementEvaluator#join_array_param` deliberately contributes nothing for index-write forms ("no new element evidence we can cheaply attribute"), which is exactly wrong for a compound write: `t[0] += 5` manufactures a NEW value (`old + 5`) that is provably outside the retained constant union. The stale union then feeds the constant-comparison fold and fires on correct code.

For contrast, the sibling shapes are safe today: `x.attr += v` on a fold-safe struct local degrades the member read to untyped (imprecise, no wrong type), and a plain `t[0] = v` joins `v`'s type via the `[]=` arm.

**Fix bar** (the #539/#540 lesson: fix the stale fold at its root, don't blunt the diagnostic): an index COMPOUND write must either join the computed RHS result type into the element evidence (the compound machinery already types `t[0] + 5` — route that type into `array_added_elements`, same for the Hash arm's value side) or, where the result type is unavailable, widen the retained constant elements to their nominal base. Gate with the usual pair: this repro must go quiet, and a plain `t[0] = v` / `<<` case must keep its joined element precision; corpus FP diff.

zonuexe:
Implements #560 (design option 1): straight-line content mutations now JOIN the added value's type into the carrier's element evidence, killing the live false-`always-falsey` family on correct code — `u = [1, 2]; u.push(6); u.last == 6` no longer folds falsey, and the `t[0] += 5` / `t[0] = 6` / redmine `csv_options[:encoding]` shapes stay quiet with the genuine-falsey control still firing.

**Join semantics — admissibility-gated:** the mutator's argument types reach the widening seam as `arg_types:`; per added member, after value-pin widening, the value joins precisely when the seed's own class set (read off the PRE-state literal, not the widened carrier) already admits it, and contributes `Dynamic[top]` when it does not. The subtype oracle accepts a union containing `Dynamic`, which is what keeps haml's declared-sig shape quiet: on haml's real filter sources, removing the gate reproduces exactly the 8 `def.return-type-mismatch` FPs PR #561 recorded; with the gate, 0. The shared join algebra is extracted to `content_join.rb` so the block-capture (ADR-56 slice C) and straight-line paths use one copy — the ADR-56 `block_captured_writeback` guard spec is unchanged and proven live by deliberate corruption.

**The corpus gate found one FP in the first draft and its diagnosis reshaped the fix:** mail `message.rb:1841` (`hash['multipart_body'] << part.to_yaml`) fired `undefined method '<<' for Hash[…]` because mutation widening is a ONE-WAY door — the first store's join became the whole value parameter for good (`widen_for_mutator` has no `Nominal` arm), so precision from one sample of a growing population was a lie. The shipped rule is asymmetric by carrier semantics: **Array element evidence joins precisely** (a union over positions — a missed store is merely incomplete), **a widened Hash's key/value parameters stay gradual** (a read selects one key; a missed store makes it wrong). An intermediate draft that gradualized both sides regressed ADR-56's recorded slice-C win (`acc = []; acc.push(m)` fell to `Array[Dynamic | Integer]`) — the golden snapshot caught it, and that regression is the evidence the line sits in the right place. The one-way door itself (later stores invisible to a widened carrier) is pre-existing and filed as #580 with the provenance-mark design this change deliberately avoids.

**Corpus gate** (full 11-target set, base = post-#576 master, `--no-baseline --no-cache`): **+0 / −1** — the removal is kramdown `hash_ast.rb:28` (`undefined '<<' for Symbol`), confirmed as this exact family (first store pinned the value union; the join removed the stale fold). mail, redmine, mastodon, haml and the rest byte-identical. mail's ragel cluster (#533 item 8) does NOT improve — its blocker sits upstream of this seam and the issue's precision half stays open, verified before/after.

Specs: `content_join_spec` (38), the `arg_types` describe in `mutation_widening_spec`, integration fixtures pinning the exact diagnostic line sets to `# GENUINE-FALSEY` / `# GENUINE-UNDEFINED` markers (so "went quiet" cannot pass on a dead rule), the haml-shaped declared-sig fixture with its `genuinely_wrong` live-rule control, and the three `to_yaml` forms of the mail shape (block / unconditional / straight-line). Non-vacuity proven by deliberate breakage: stubbing the join, removing the admissibility gate, and corrupting the ADR-56 guard each turn the named examples red. One deliberate ADR-14 deviation: `mutation_join_declared_sig/sig/temple.rbs` is hand-written because it models haml's *imprecise* third-party signature — `sig-gen` would emit the correct inferred type and destroy the test (same pattern as `user_methods_with_sig/sig/parity.rbs`).

ADR-56 carries the WD2.5 note for the moved slice-C names and the two straight-line gates. Full `make verify` exit 0 on the rebased tree (post-#579).

Closes #560.


zonuexe:
Landing gates, all green:

- **`make verify`**: exit 0 (run on the B1-fix head and again on the B2-fix head; the final commit is the ADR correction, docs-only, `make docs-check` exit 0).
- **Corpus FP diff** (full 11-target set, base = post-#576 master, `--no-baseline --no-cache`, re-collected after EACH review round): **+0 / −1** every time — the single removal is kramdown `hash_ast.rb:28` (`undefined '<<' for Symbol`), confirmed as this exact stale-fold family. Zero diagnostic movement anywhere else; the widened unions move type strings only.
- **Critical review** (independent, three rounds): R1 found B1 — the empty-seed precise Array join + the one-way widening door produced a NEW wrong-type dispatch FP (`a=[]; a.push(1); a.push("s"); a.last.upcase` → error; master quiet) — fixed by extending the gradual floor to the straight-line Array side and reframing the design around sees-one-store (straight-line) vs sees-all-stores (block path). R2 found B2 — the fix's `drop_dynamic` Union-flattening dropped DECLARED gradual arms (`Array[Integer | untyped]` closed and fired) — fixed at the true source: `grep_v` restored, `loop_content_writeback` seeds non-rebound names from the pre-body scope. R3: every probe re-verified (zero new, four removals vs master, controls firing), final condition was the ADR-56 paragraph still describing the abandoned flatten — corrected in `b42864ad` with the false path recorded. **APPROVE.**
- Review by-products filed: #586 (pre-existing: declared `Array[untyped]` closes under block mutation, fires on both heads), #587 (pre-existing block-return wrong-precise families), #580 updated (the residual is NOT merely imprecise — the provenance mark is the precision-restoring follow-up).

PR #582, opened 2026-09-01
Read the model_index fact as the Hash it publishes in rigor-shoulda-matchers

### LINKED ISSUE #573: rigor-shoulda-matchers reads the :model_index fact as a ModelIndex object, but the fact publishes a Hash

zonuexe:
Fixes #573. The break was real and worse than filed: `rigor-shoulda-matchers`' analyzer consumed the ADR-9 `:model_index` fact as a `ModelIndex` OBJECT (`index.find(anchor)` → `entry.column?`), but the fact publishes a plain Hash — `Hash#find` with one non-block argument returns an Enumerator, so on any real project with both plugins active every shoulda-matchers file produced `plugin "shoulda-matchers" raised during diagnostics_for_file: NoMethodError: undefined method 'column?' for an instance of Enumerator` — one noisy `runtime-error` diagnostic per file replacing every real check (verified end-to-end by reverting the analyzer on a two-plugin fixture project). The plugin's own spec stubbed a ModelIndex-shaped object for the fact, so the gate could not see the published shape — the same green-gate-that-cannot-see class as the doc-extractor postmortem.

**Fix direction: the consumer, not the producer.** `docs/internal-spec/plugin.md` states the fact-store contract — the value is data, not objects — and the two other consumers (`rigor-actionpack`, `rigor-factorybot`) already index the Hash correctly (`model_index[model_class]` → `entry[:columns]`). The analyzer now does the same; publishing the object instead would have broken both working consumers and the contract.

**The spec gap is closed the way the issue demanded:** the stub lets are now the real published Hash shape, and a new cross-plugin integration block builds a real tmpdir project (`db/schema.rb` + AR model + spec file), runs BOTH plugins through the actual Runner, and pins must-fire `unknown-column` on `have_db_column(:nonexistent)` + must-not-fire on a real column — each also asserting no `runtime-error` leaked. Regression-proven: with the pre-fix analyzer swapped back in, the must-fire case fails exactly as the broken path predicts.

Corpus FP diff deliberately not run: neither survey app's config loads rigor-shoulda-matchers, so the diagnostic arms cannot move by construction. Full `make verify` exit 0 on the rebased tree; the other consumers' spec files re-run green (actionpack 54, activerecord 127, factorybot 19).

Closes #573.


zonuexe:
Landing gates, all green:

- **`make verify`**: exit 0 on the rebased tree (the review-round commit touched one spec file + one comment; the spec file re-run green, 21/21, and CI covers the full suite on the final head).
- **Corpus FP diff**: deliberately not run — neither survey app's config loads rigor-shoulda-matchers, so the diagnostic arms cannot move by construction.
- **Critical review** (independent): REQUEST-CHANGES with one verified blocker — the new integration spec's `runtime-error` guards filtered on the plugin's own `source_family` while the raise envelope is stamped `:plugin_loader`, so they could never fail (the exact blind-gate class this PR exists to close). Fixed in `fd5c53f5` (assert over the unfiltered diagnostics) and swap-PROVEN to discriminate: with the pre-fix analyzer restored, the must-not-fire example now fails on exactly that assertion. Stub association rows aligned to the producer's published shape (`polymorphic:` key, nil polymorphic target) and the analyzer's shape comment corrected in the same commit. The reviewer's remaining find — `class ::User` models keyed `"::User"` miss ALL three Hash consumers, a uniform pre-existing FN — is filed as #583.

PR #584, opened 2026-09-01
Thread scope through multi-statement block bodies in the block-return pass

zonuexe:
Implements item 9 of #533: the block-return-typing pass evaluated only the block body's last statement in the entry scope, so a tail reading a block-local answered Dynamic (`m.synchronize do v = 42; v end`) while `{ 42 }` bound. `type_block_body` now runs a gated scope-threading fold via the main pass's own `StatementEvaluator` (`Scope#evaluate`), falling back to the tail-only path on: non-StatementsNode bodies, single-statement bodies (the common case, zero new cost), re-entrancy (no fold inside a fold — the ADR-56 BodyFixpoint already re-evaluates nested block bodies), and tails that read no name an earlier statement writes (two cheap DFS passes; measured isolation shows the scans are free and 100% of the ~+2% CPU delta is gated re-evaluations doing real work).

Controls (all pinned as specs, non-vacuity proven by reverting the engine file — exactly the 10 fire-examples go red, all 6 unchanged-controls stay green): `do v = 42; v end` → 42, `[1,2].map do q = 42; q end` → [42, 42], def-param tails stay honest-Dynamic, outer-local/begin/conditional/early-return semantics unchanged, `arr = []; arr << 1; arr` widens instead of folding back (why the fold reuses StatementEvaluator rather than a hand-rolled prefix walk — a hand-rolled one would re-introduce the ADR-56 unsoundness class). Both corpus beneficiaries improve: the concurrent-ruby read_write_lock shape (bool → true) and the textbringer lsp/client shape (Dynamic → 1). Alternating-arm diagnostic A/B over all of lib/ + the check-plugins surface: byte-identical, ~70k lines.

Draft while queued for the serial landing slot (corpus arms + review); remote CI = the post-rebase verification per the updated flow.

Refs #533.


zonuexe:
Landing gates, all green:

- **Corpus FP diff** (full 11-target set, base = post-#581 arms, `--no-baseline --no-cache`): **+0 / −0** — the threading adds precision without moving a single diagnostic.
- **Critical review** (independent, two rounds): R1 REQUEST-CHANGES — a value-carrying `next` before a gated tail turned honest Dynamic into a wrong-precise fall-through value; fixed with a boundary-aware jump scan (`prefix_statement_jump_free?`: writes count at any depth, jumps only above the nearest retargeting construct), plus the per-element threading arity cap (8, shared with the Range limit) and the count-pinned dump assertions. R2 **APPROVE** — and the reviewer traced their own contradicted `break` claim to a cwd-contaminated compound-shell measurement, confirming the author's A/B (master answers Dynamic for ALL five jump shapes; pinning `Dynamic[top]` for the break sibling is the only sound answer). Non-blocking review by-products: #587 (two pre-existing block-return wrong-precise families this PR neither caused nor fixes).
- Local `make verify` green pre-review; remote CI green on the rebased final head (per the draft-CI flow, the post-rebase verification).

PR #585, opened 2026-09-01
Type the Rails singleton readers, Duration multipliers, and sidekiq jids

zonuexe:
Implements items 2-4 of #534 (item 1 landed as #578; item 5 + AMS stay open on the issue), each FP-adjudicated by measurement:

**Rails singleton readers** (`rigor-railties`, which already claims this territory for effects): `Rails.logger` → `ActiveSupport::BroadcastLogger`, `.cache` → `ActiveSupport::Cache::Store`, `.configuration` → `Rails::Application::Configuration`, `.application` → `Rails::Application` — via a syntactic receiver gate (literal `Rails`/`::Rails`, no args/block; the constant is unresolved so a type-side gate cannot work). The issue's suggested `logger -> Logger` is the one WRONG choice: stdlib `::Logger` is RBS-known, the leniency gate stops declining, and working Rails code draws 4/4 measured FPs (`tagged`/`silence`/`broadcast_to`/`local_level=`) — pinned as a red-under-that-arm spec. Defensive idioms (`Rails.logger || fallback`, `return unless app`, `.nil?` guards) measured silent.

**Duration multipliers** (`rigor-activesupport-core-ext`): the RBS `-> untyped` STAYS (declaring `ActiveSupport::Duration` in RBS would make it a known-but-incomplete class → every method_missing member a false undefined-method); a `dynamic_return` supplies the `Duration` nominal. The important find: the multiplier alone REGRESSED five shapes — with a named class the overload selector commits wrong arms (`Time.now - 30.minutes` → Float, `2 * 1.day` → Integer, `Date.today - 1.week` → Rational; 5/5 measured downstream FPs) — so a second `dynamic_return` on `+ - *` ships as the correction (Time/DateTime keep their kind; numeric×duration → Duration; `Date ± duration` → `Date | Time`; `/` and duration×duration declined). Plain arithmetic proven untouched (`1 + 1` → 2, Time−Time → Float, own `#days` unhijacked). Perf A/B: inside noise. One intended side effect, disclosed: the plugin's Duration effect rows now MATCH (`7.days.ago` attributes `[global.read, nondet.time]` — the rows were dead against a Dynamic receiver), so a project with a tight effects envelope could newly exceed it.

**sidekiq jids** (`rigor-sidekiq`): `perform_async`/`perform_in`/`perform_at` on a DISCOVERED worker → `String`; `perform_inline` excluded (returns `#perform`'s value). The complete reading `String?` (client middleware can halt → nil) was measured: 1 error-severity possible-nil on the assigned shape with per-method unpredictability, vs 0 diagnostics for `String` across every fixture shape incl. nil-guards — FP-first picks `String`; both arms pinned, the rejected one verified red. Non-worker `perform_async` keeps its own answer.

Combined idiomatic three-plugin fixture: byte-identical diagnostics before/after. Draft while queued for the serial landing slot (corpus arms on redmine/mastodon + review); remote CI = the verification per the updated flow.

Refs #534.


zonuexe:
Landing gates, all green:

- **Corpus FP diff** (redmine + mastodon — mastodon's config loads rigor-activesupport-core-ext, so the Duration multiplier and arithmetic rules run in anger there; base = post-#581 arms, `--no-baseline --no-cache`): **+0 / −0** on both.
- **Critical review** (independent): **APPROVE**, no blocking findings. The reviewer independently re-verified all three FP adjudications (the `::Logger` arm's 4 FPs reproduced exactly; the Duration corrective sweep incl. non-literal operands, chained arithmetic, kwarg positions, and a user-defined `+` left intact; the sidekiq `String` answer silent across five untested guard shapes with the protection genuinely active), the effects side-effect (labels ADDED only, snapshot drift legible, `effect.envelope-exceeded` not newly firing), and the version-bump/cache-key story (the two `incremental_state_fingerprint` sentinels are what keep the plugins out of the ADR-88 opaque lane). Three non-blocking follow-ups filed as #588 (comment drift, `duration + Time` typing crashing code, the project-defined-`Rails` shadow — verified diagnostics-silent).
- Remote CI green on the rebased final head (draft-CI flow).

Issue #590, opened 2026-09-01
call.unresolved-toplevel fires on struct-factory block bodies' own member reads


zonuexe:
Noticed during #525's `:self` sentinel work (pre-existing, unchanged by it): the main walk over a struct-factory block body still emits `call.unresolved-toplevel` for the body's own member reads — `Struct.new(:text) do def shout; text.upcase; end end` warns "unresolved toplevel call to 'text'" even though the member read is genuine and (post-#525-branch) types correctly on the re-typing path. The block-def re-homing (#555) registered the defs but the WALK's unresolved-toplevel rule still evaluates the body in the enclosing (toplevel/module) context, so every member accessor read looks like an unknown toplevel call — diagnostic noise on every struct-factory body, mail/haml/faraday shaped.

Fix direction: the rule should recognise a body whose enclosing factory carries `struct_member_layouts` members (the same table #555 seeds) and treat member names as resolved there — or the walk should re-home the body's self the way the re-typing path does. Must-fire control: a genuinely unknown name inside the same block body still warns.

Issue #589, opened 2026-09-01
A while loop erases a struct local's carrier even when the body never touches it — the real blocker under mail's ragel cluster


zonuexe:
Found implementing #525's setter-rejoin half (branch `struct-factory-self-sentinel`): the premise that "a setter retracts fold-safety, after which reads decline" does not hold — the actual blocker for mail's ragel cluster (~250 sites, #525's re-measured evidence) is upstream: **any `while` loop erases a struct local's `StructInstance` carrier, whether or not the loop mentions the struct.** Measured: `s = AddressStruct.new(...); while cond; i += 1; end; s.raw` answers Dynamic — the carrier is gone before member folding is ever consulted; the straight-line setter shape (`s.local = chars(d); s.raw`) already works today.

So on `lib/mail/parsers/address_lists_parser.rb` — one giant ragel `while` state machine — no fold-safety improvement can pay: the loop-seam widening (ADR-56 / MutationWidening territory, where the #560/#581 added-value join work lives) drops the carrier wholesale instead of widening only what the loop body's evidence touches.

Fix direction: the loop exit scope should preserve a local's carrier when the loop body neither rebinds nor content-mutates that name (the same does-the-body-touch-it analysis `loop_content_writeback` already performs post-#581 — its non-rebound-name arm reads the pre-body scope; the carrier-typed locals seem to take a different, erasing path — find where and align them). Must-fire control: the mail shape above folds `s.raw` after an untouching `while`; must-still-widen: a loop that DOES setter-write the struct keeps declining (the `deferred_setter` gate is load-bearing — with it removed, `[1].each { s.local = "L" }; s.local` answers a stale `nil`, verified unsound).

Gate: corpus FP diff (mail is the target — its ragel precision is the payoff), ADR-56 fixtures byte-green, `make verify`.

zonuexe:
Mechanism correction from the implementation (PR #596): **the carrier was never erased** — after an untouching `while`, the local still reads its full `StructInstance`; what died was the fold GRANT: `Scope#join` omitted `struct_fold_safe_locals` from its constructor call, falling back to the empty default, so every merge in a method body (an `if` exactly as much as a `while` — verified) silently revoked struct folding for everything after it. PR #596 carries the set through joins by intersection.

[...]

Issue #525, opened 2026-08-31
Struct.new factory classes lose members, .new, and (with a block) self scoping


zonuexe:
Found by the 2026-09-01 corpus opacity sweep on six targets (branch `opacity-sweep-20260901`: reports for haml, hamlit, faraday, mail, tdiary-core, algorithms). Three stacked shapes, worst first:

1. `X = Struct.new(:a, :b) do ... end` — the do-block defeats ADR-48 folding entirely: `.new`'s result, member accessors, and the block-defined methods all answer Dynamic, and `self` inside the block's `def`s types as the ENCLOSING module (faraday: `singleton(Faraday)` inside `Request`'s body, `lib/faraday/request.rb:27`), so the whole Options/Request layer misses. haml/hamlit `ParseNode`/`Line` ~285 sites; repro `repros/typeof_probe9.rb`.
2. `AddressStruct = Struct.new(:a, :b)` (no block, constant-assigned factory) — mail reports the factory call AND `.new` on the constant both Dynamic (39 sites); tdiary-core `Market = Struct.new(:host, :region)` same-file (18). The blockless factory is supposed to be ADR-48's covered case, so part of this may be the #513 under-seeded lens — re-verify against `rigor check` before scoping.
3. `class Deque < Struct.new(:x)` subclassing (algorithms — several hundred cascade sites through `left`/`right`/`key` chains).

**Fix direction:** treat the `Struct.new(...) do ... end` block as a class body for the synthesized struct class (self = the struct singleton inside `def self.`, instance for `def`), and register block-defined methods in the discovered tables under the factory's assigned constant; the member layout machinery (`struct_member_layouts`) already exists and is seeded by the check walk. ready-for-human: the block-body re-homing touches ScopeIndexer's class-body model, which has ADR-17/ADR-43 interactions worth a design pass.

**Gate:** `make verify` + self-check gates; corpus FP diff; must-still-succeed controls for the plain `Struct.new` fold.

zonuexe:
Design for the in-body member-read residual (diagnosed 2026-09-01, on the all-in integration with #555):

[...]

PR #591, opened 2026-09-01
Grant struct-member folding to in-body reads through the :self sentinel

zonuexe:
Implements #525's remaining residual: in-body member reads of struct-factory block defs now fold when the CALL SITE's receiver is fold-safe — `Line.new("a", 2).shout` (body `text.upcase`) went `Dynamic[top]` → `"A"`. The design is the issue's `:self` sentinel (collision-free in `struct_fold_safe_locals`, computed at `try_user_method_inference` from the same two tests `foldable_receiver?` applies, carried into the ADR-84 return-memo key so key and scope cannot disagree, with `RecursionContext` carrying the polarity for the bot-collapse retry).

**One deliberate deviation from the issue's design, with the counter-example recorded:** guard (c) as specified ("no member setter on self") is insufficient — `def reset!; self.text = ""; end / def shout; reset!; text.upcase; end` would fold `text` to the construction value while the runtime read is `""`. Unrecognised self-calls therefore refuse — and to keep the design's propagation arm alive (a delegating body inheriting the grant), self-calls resolve against the receiver's own class and are asked the same question recursively, cycle-guarded and depth-capped at 4. `def outer; shout; end` keeps the grant; anything reaching a writer at any depth loses it. Discrimination proven by breakage: deleting the memo-key element serves one call site's fold to a receiver nothing proved current — unsound, and the spec catches it in both memo orders.

**The setter-rejoin half (issue comment 2) is deliberately NOT here, and its premise is refuted by measurement:** the straight-line setter shape already works on master (`s.local = chars(d); s.raw` → "r"), and the real blocker under mail's ragel cluster is that **any `while` loop erases a struct local's carrier even when the body never touches it** — an ADR-56 loop-seam issue filed as #589 with the measured table; a fold-safety rejoin cannot pay those sites. Also filed from this work: #590 (`call.unresolved-toplevel` noise on struct-factory bodies' own member reads, pre-existing).

Specs: 81 examples across the two engine spec files — must-fold (implicit-self, explicit `self.member`, delegation, two call sites folding to their OWN carriers), must-decline (own setter, sibling mutator, two-hop transitive, mutual recursion + non-hang control, self-escape, non-foldable call site), memo polarity in both orders, and the unmoved floor. Data control unchanged (`Data.define` body folds as before — Data needs none of this). `spec/rigor/inference` (2967) + `spec/rigor/analysis` (1382) green — run because the memo-key change touches every user-method inference.

Draft pending the landing gates (full 11-target corpus arms + independent review + CI per the pipeline).

Refs #525.


zonuexe:
Landing gates, all green:

- **Corpus FP diff** (full 11-target set, base = current-master arms, `--no-baseline --no-cache`, re-collected on the fixed head): **+0 / −0** — the grant machinery adds precision (`Line.new(...).shout` folds its own member value) without moving a diagnostic.
- **Critical review** (independent, two rounds): R1 REQUEST-CHANGES — the guard machinery survived every evasion probe fail-closed (send/public_send, aliases in three spellings, block- and lambda-mediated writers, `instance_variable_set`, depth-cap and cycle bounds), but the caller-side "any chained call is fresh" arm served construction values on fluent builders (`Line.new("a").with_text("z").shout` → "A", runtime "Z"). Fixed: the arm now recognises MATERIALISATIONS only (`.new`/`.[]` on a layout-known constant, `.with` only when no hand-written `with` exists — the override guard verified empirically). R2: **APPROVE** — whitelist verified with the reviewer's own probes incl. the hand-written-`with` evasion, the memoization's outer/inner boundary confirmed structurally sound, the 4-hop cap empirically exact at the boundary, ADR-48 accurate. Pre-existing twin filed as #595 (slice-2 `fresh_receiver?` trusting self-returning chains — with the reviewer's composed d3 shape recorded there); the ADR now carries the link.
- The deliberate design deviation from the issue (recursive self-call grant, cycle-guarded and depth-capped fail-closed, replacing the insufficient "no member setter on self" guard) is recorded in the ADR-48 Slice-5 note with its counter-example.
- CI: green on the final head. The setter-rejoin half of #525 was measured OUT (the real blocker is #589 — a `while` loop erasing struct carriers wholesale); #525 stays open on that residual.

Issue #542, opened 2026-08-31
Hash.new with a default value or block still reads V? through RBS Hash#[] — default-bearing hashes never return nil


zonuexe:
Found adjudicating #518/#519/#540's corpus gate. textbringer's ctags builds `tags = Hash.new { |h, k| h[k] = [] }`; every `tags[name]` read is non-nil at runtime by construction, but the engine types the receiver as a plain Hash and RBS `Hash#[] -> V?` supplies a nil arm, so downstream `candidates.size` / `candidates[index]` fire `call.possible-nil-receiver` on a working program (2 sites, `rigor-survey/textbringer/lib/textbringer/commands/ctags.rb:30,47`) once the surrounding types become honest (#540 exposed them).

Fix shape needs a type-model decision (hence ready-for-human): a carrier property "default-bearing Hash" set by the `Hash.new(arg)` / `Hash.new { }` constructor fold and consulted by the `[]` dispatch to drop the optional — a small ADR-1/type-spec addition, since carriers are normative in `docs/type-specification/`. The property must survive local flow (assignment, return through an inferred method) to pay off; deciding how far it propagates is the design question. A cheaper sibling: flow narrowing "a completed method call on x implies x was non-nil" would ALSO silence these two sites (the code calls `candidates.empty?` before the flagged reads) and kills a broader FP class, with its own reachability questions.

Evidence preserved in the #518 corpus-gate adjudication (PR body) and branch `opacity-sweep-harness-20260901`.

PR #592, opened 2026-09-01
Type the ActionDispatch request predicates, format, and flash chain

zonuexe:
The last agent-adjudicable slice of #534's request-surface lane, every contract read off rails v8.0.2 / rack v3.1.8 source:

**15 request predicates → `bool`** (post?/get?/…/xhr?/ssl?/local?/form_data?…): admission rule = zero-argument predicate whose body cannot return a non-boolean. The central measurement: a plain `bool` does NOT fold (`true | false` never proves one constant) — `return unless request.post?`, the `request.get? ? :read : :write` ternary with a live guard (the exact shape that rejected `Parameters#[]` in #578), `&&`/`||`/`!` all draw zero diagnostics; typing the predicates `Constant[true]` instead turns the fold spec red with 4 firings, so the union is doing the work.

**`request.format` → `Mime::Type | Mime::NullType`** — the union IS the contract (`formats.first || Mime::NullType.instance`, and the empty case is reachable via `?format=bogus`); both arms RBS-less → lenient.

**Flash chain**: `flash.now` → `FlashNow` (never nil), `keep`/`discard` → `FlashHash` arity-gated to the zero-argument spelling (`k ? self[k] : self` — `keep(:notice)` correctly declines).

**Measured non-changes, recorded in the plugin comment + manual:** `flash[:k] = v` / `session[:k] = v` need NOTHING (assignment-expression semantics already type them; `--protection` reads 6/6 with no plugin rule — the corpus's 127+29 pair counts are dispatch-resolution counts no return contribution can move). `flash[:k]` / `session[:k]` READS stay untyped: both candidate arms were run and each draws its own FP (non-nil folds the live guard; `String | nil` fires error-severity possible-nil on `note.upcase`) — `session[:user_id]` is THE nil-checked read in Rails, so this lands squarely on #574's blocker, pinned as paired arm specs.

End-to-end `SessionsController` fixture exercising every touched surface: byte-identical "No diagnostics" before and after. Out of scope, recorded: `Mime::Type` predicate methods (open method set — needs its own adjudication), nullable String readers (`user_agent`/`referer` reopen the #574 story), `Session#key?` (near-zero sites).

Draft pending the landing gates (corpus arms on redmine/mastodon + review + CI).

Refs #534.


zonuexe:
Landing gates, all green:

- **Corpus FP diff** (redmine + mastodon, base = current-master arms, `--no-baseline --no-cache`): **+0 / −0** on both.
- **Critical review** (independent, two rounds): R1 REQUEST-CHANGES — the `request.format` union was proven not flow-inert (ternary arm-folding on a live guard) and its justifying comment carried two false claims (`Mime::NullType#nil?` is an explicit `true` — a nil-masquerading object a nil-free union mismodels, with mastodon branching on `request.format.nil?` in production). Adjudication: the format surface is DROPPED from this PR and filed as #594 (a type-model question — nil-masquerading objects — adjacent to #542's carrier-property pattern); the withdrawal is PINNED as two specs proven red under restoration. Version corrected 0.10.0 → 1.0.0 per the single-digit rule (a manifest version is a cache key, not a gem release). R2: **APPROVE** — withdrawal verified dropped and genuinely pinned, repaired negatives non-vacuous, and the narrowed inertness comment's causal claim (`.nil?` folds on RBS-known non-nil nominals, not on RBS-less ones) independently verified.
- What lands: 15 request predicates → `bool` (the central measurement: a two-constant union cannot arm-fold — the `Constant[true]` arm turns the fold spec red with 4 firings) + the FlashHash chain (`now`, arity-gated `keep`/`discard`), plus the recorded measured non-changes (`[]=` needs nothing; flash/session reads stay untyped, both arms FP-measured — #574's case, further evidenced).

PR #593, opened 2026-09-01
Widen retained constants when a mutator's arguments cannot be read

zonuexe:
Fixes residual (a) from the PR #581 review, recorded on #580: a content mutator whose arguments the extractor cannot read (`m = [1, 2]; m.concat(xs)`) used to keep the seed's constant union untouched, so `m.last == 6` still folded always-falsey on correct code — the same falsified-evidence family #560 closed for readable-arg mutators. `unpinned_elements` now widens the retained constants to their nominal base exactly when `arg_types` is non-empty while extracted evidence is empty (real arguments the extractor could not read); an EMPTY `arg_types` still means the block path, whose slice-C re-derivation re-adds evidence itself, so its seed pinning is untouched.

Paired specs (the widening + the block-path non-change) plus the `concat_unknown` integration fixture with the genuine-falsey control still firing. ADR-56 gains WD2.6 covering only what ships — including the recorded pointer that a scope-side provenance mark cannot carry protection across a method return (the #580 exploration's measured wall; evidence on the issue).

Draft pending the landing gates (full 11-target corpus arms + review + CI).

Refs #580.


zonuexe:
Landing gates, all green:

- **Corpus FP diff** (full 11-target set, base = current-master arms, `--no-baseline --no-cache`, re-collected after the fix round): **+0 / −0**.
- **Critical review** (independent, two rounds): R1 REQUEST-CHANGES — the first cut CLOSED the element parameter on zero evidence, violating the module's own one-store invariant, with two measured FPs (a sig'd unreadable `concat` drew the exact `Array[Symbol]` return-mismatch #561 scoped away; a post-concat dispatch read errored on the closed union) that the corpus could not see. Fixed by routing the unreadable store through the ordinary one-store pipeline (`join_array_content` + `gradual_floor`) → `Array[1 | 2 | Dynamic[top]]`; both regression surfaces pinned as fixtures proven red-under-restoration; the `concat([])` lost-fold recorded as a deliberate FN (a no-op-call FN beats the two FPs). R2: engine APPROVED; final condition was WD2.6's heading still describing the reverted cut — corrected. The review also surfaced the family's third door (in-block unreadable args, pre-existing on both arms), recorded on #580.
- What lands: `m = [1, 2]; m.concat(xs); m.last == 6` no longer folds always-falsey on correct code, with seed pinning intact so declared-sig acceptance is unchanged.

PR #596, opened 2026-09-01
Carry the struct fold-safety set through scope joins

### LINKED ISSUE #589: A while loop erases a struct local's carrier even when the body never touches it — the real blocker under mail's ragel cluster

zonuexe:
Implements the mechanism behind #589 — which the investigation corrected: the carrier was never erased. `Scope#join` omitted `struct_fold_safe_locals` from its constructor call, falling back to the empty default, so EVERY merge in a method body — an `if` exactly as much as the issue's `while` — silently revoked the fold grant for everything after it. The fix intersects the set across the merge (both arms normally carry the identical statically-scanned set; intersection is the FP-safe direction since the grant licenses a fold). Nothing about fold SAFETY moved: the static scan (deferred setters, rebinds, escapes) is unchanged — the join was discarding its answer, not contradicting it.

This is also the missing half of #591's flat corpus effect: the grant machinery was revoked by the first branch in most methods; with the set carried through joins, fold-safe struct locals survive ordinary control flow (verified: reads fold after an untouching `while`, after an `if`, and inside a LATER loop across the ADR-56 fixpoint; a slice-4 member write-back now survives a merge).

**The measured negative result, recorded on the issue: mail's ragel cluster does NOT unlock**, for reasons this fix correctly leaves alone — its 131 in-loop `address.<member> =` setters hit the load-bearing `deferred_setter` gate (removing it serves a stale nil, proven), and `address_list` escapes via bare returns. That residual is filed separately as the per-iteration setter-effect question.

Non-vacuity: removing the one join line turns exactly the three must-fold examples red, leaves the three must-decline green (deferred-setter-in-loop, rebinding loop, escapes). Controls live as heredocs (the fixture auto-formatter strips 'useless' if guards — every shape here needs one). Corpus spot-set (haml, kramdown converters, mail parsers) byte-identical; the full-11 arms run at landing with specific attention to new flow folds / return-mismatches on struct-heavy targets (the fix can only GRANT folds the static scan approves — any bug it exposes is a pre-existing scan bug previously masked by branching).

Closes #589.

### DIFF
--- a/changelog.d/fixed/loop-preserves-untouched-carriers.md
+++ b/changelog.d/fixed/loop-preserves-untouched-carriers.md
@@ -1,0 +1,1 @@
+- **[engine]** A `Struct` member read now still folds to its known value after an `if` or `while` earlier in the method, instead of going untyped for the rest of the body whenever the code branched at all. ([#589](https://github.com/rigortype/rigor/issues/589))
--- a/changelog.d/fixed/loop-preserves-untouched-carriers.md
+++ b/changelog.d/fixed/loop-preserves-untouched-carriers.md
@@ -1,1 +1,1 @@
-- **[engine]** A `Struct` member read now still folds to its known value after an `if` or `while` earlier in the method, instead of going untyped for the rest of the body whenever the code branched at all. ([#589](https://github.com/rigortype/rigor/issues/589))
+- **[engine]** A `Struct` member read now still folds to its known value after an `if` or `while` earlier in the method, instead of going untyped for the rest of the body whenever the code branched at all. ([#589](https://github.com/rigortype/rigor/issues/589), [#596](https://github.com/rigortype/rigor/pull/596))
--- a/docs/adr/56-block-captured-local-mutation.md
+++ b/docs/adr/56-block-captured-local-mutation.md
@@ -370,6 +370,19 @@ that scan's answer rather than contradicting it. All three still
 decline, and folding now also holds across this ADR's loop fixpoint
 rather than for a single pass.
 
+Restoring the grant did expose a real gap in that scan, fixed in the
+same change. The scan's counting identity is about the LOCAL and says
+nothing about a member read's RESULT: `s.x << v` mutates the container
+`s.x` returns while `s.x` is a textbook pure read, so the local stayed
+fold-safe while its member's value changed underneath. A local whose
+member-read result is itself a receiver is now disqualified outright,
+with no allow-list of its own — `s.x.to_s` loses precision for
+nothing, but an allow-list is what produced the bug, and being too
+broad only costs a `Dynamic[top]`. That also removes a pre-existing
+false positive on the straight-line form, which fired before this
+branch existed. The scan header's claim that a missed case is "never
+unsound" was false and is corrected there. Remaining residual: #597.
+
 **The payoff was measured and it is NOT mail's ragel cluster**, which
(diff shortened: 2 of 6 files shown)


zonuexe:
Landing gates, all green:

- **Corpus FP diff** (full 11-target set, base = current-master arms, `--no-baseline --no-cache`, re-collected after the fix round): **+0 / −0** both rounds.
- **Critical review** (independent, two rounds): R1 REQUEST-CHANGES — the join mechanics survived every probe (conditional rebind/setter/escape, multi-assign, rescue/ensure/case seams, nested joins), but the newly-reachable fold surface exposed a pre-existing scan gap: `s.x << v` counted as a pure read of `s`, so post-branch reads served the stale container (`undefined method 'upcase' for 5` on correct code). Fixed by disqualifying a local on ANY call chained off a member-read result (no allow-list — an allow-list is what produced the bug; over-breadth costs only Dynamic), which also cures the pre-existing straight-line base FP; the argument-position counter-edge (member reads as arguments must keep folding) is pinned. R2: **APPROVE** — all probes closed, the chain/argument distinction verified, the rubocop-driven restructure drift-free, and the one residual (assignment-capture aliasing) PROVEN engine-wide policy by a plain-local control — this PR restores parity with the engine's binding model rather than adding optimism. That family is now umbrella-filed as #601; the review also surfaced #600 (`opaque_block_self` dropped at joins, same bug class, pre-existing).
- The corrected mechanism is recorded on #589 (the carrier was never erased — `Scope#join` dropped the fold-safety set; an `if` revoked exactly as the issue's `while` did), with the measured negative result (mail's ragel cluster does NOT unlock — its 131 in-loop setters hit the load-bearing `deferred_setter` gate; successor filed as #597) and the instrument note (`type-scan` cannot see Dynamic→precise). This also explains #591's flat corpus effect: the grant died at the first branch of most methods.

Issue #293, opened 2026-08-07
Nested Struct member pinned to its empty construction literal folds .first to nil and fires on correct code


zonuexe:
> *This was generated by AI during triage.*

## What to build

Reported by the rigor-rs differential harness (upstream-feedback batch 2, `20260807-upstream-feedback-batch2.md`), bisected on their side to the #271 fix and reproduced here on master (`dfdced55`).

Typing a nested `Struct.new` constant cross-file (the #271 fix) also lets the constructor call pin a member to its **construction-time literal**. When that literal is the empty array, `.first` folds to `nil` and the next call fires `undefined method ... for nil` — on correct code whose array is filled by `<<` immediately after construction.

Minimal repro (2 files, fresh cwd; `rigor check lib.rb use.rb`):

```ruby
# lib.rb
module Pkg
  class Parser
    Result = Struct.new(:items, :errors)
    def self.parse(text)
      r = Result.new([], [])
      text.split(",").each { |t| r.items << Item.new(t) }
      r
    end
  end
  class Item
    def initialize(name) = @name = name
    def local = @name
  end
end
# use.rb
p Pkg::Parser.parse("a,b").items.first.local
```

| engine | verdict |
|---|---|
| v0.3.1 | silent |
| master (`dfdced55`) | `use.rb:1:40: error: undefined method 'local' for nil` |

The rigor-rs survey found this is one of only two diagnostics that 150 commits (v0.3.1 → `80aaf9bc`) move across 9,204 swept files, and it is a false positive. A live instance in the wild: mail's `spec/mail/parsers/address_lists_parser_spec.rb:32`.

## Agent Brief

**Category:** bug
**Summary:** Widen (or refuse to fold `.first`/`.last` on) a literal-pinned Struct/Data member that is mutated after construction.

**Current behavior:**
`Result.new([], [])` pins the `items` member to the empty array literal. A later `r.items << x` does not retract the pin, so `.first` on the member folds to `nil` and downstream calls fire.

**Desired behavior:**
A member pinned to a construction-time literal that is subsequently mutated (e.g. `<<` through the reader) is widened, the same mutation-widening polarity Rigor already applies to locals (mutators retract the `non-empty-array` refinement). The repro above must be silent; a member that is never mutated after construction keeps its precise fold.

**Key interfaces:**
- The Struct/Data member typing introduced for cross-file nested constant resolution (#271) — the pin-to-literal step needs a mutation-awareness gate, not a revert.
[...]

PR #598, opened 2026-09-01
Trust only materialisations as fresh struct receivers

### LINKED ISSUE #595: fresh_receiver? trusts any chained call: self-returning methods serve construction-time member values after mutation
Found by PR #591's review (fixtures f11/f15 in its notes), pre-existing on master with zero #591 code involved: `StructFolding#fresh_receiver?` treats ANY `CallNode` receiver as fresh ("nothing can have mutated it between materialisation and the read"), and a SELF-RETURNING method falsifies the premise:

```ruby
Line = Struct.new(:text) do
  def with_text(v) = (self.text = v; self)
  def dup_self = self
end
x = Line.new("a")
x.text = "z"          # or the fluent spelling below
x.dup_self.text        # folds "a" — runtime "z"
Line.new("a").with_text("z").text   # folds "a" — runtime "z"
```

The chained call returns the SAME long-lived object, post-mutation, but the member read folds the construction-time value — a wrong precise value on ordinary fluent-builder Ruby (the faraday Options family). The root is two-fold: `infer_user_method_return` types the body under a fixed `self_type` carrier, so `self.text = v` never poisons the returned `self`; and the freshness test never consults the callee's return. #591 tightened ITS caller-side arm to a materialization whitelist (`.new`/`.[]` on the struct class); this issue is the slice-2 direct-read twin.

Fix direction: the freshness test must refuse a chained receiver whose callee can return a non-fresh object — cheapest sound version: whitelist materializations exactly as #591's arm now does (any other callee refuses); a richer version consults the callee's inferred return for a self-alias (the #591 grant resolver walks sibling bodies already and could answer "returns self?"). Must-fire control: `Line.new("a").text` keeps folding; must-stay-Dynamic: both shapes above.

Gate: corpus FP diff; the #591 spec suite byte-green.

zonuexe:
Fixes #595: slice-2's `fresh_receiver?` read ANY chained call as transient, so a self-returning method served construction-time member values after mutation — `Line.new("a").with_text("z").text` folded `"a"` (runtime `"z"`), and the composed shape recorded on the issue (`x.dup_self.with(indent: 9).shout`) folded through the same trust. The freshness test now calls the SAME materialisation predicate #591 introduced, extracted to `MethodDispatcher::StructMaterialization` so there is one implementation (StructFolding extends it; the grant arm calls it). All three ADR-48 definition forms stay materialisations — the constant, the factory-local (`c = Struct.new(:x); c.new(1).x`, read via `scope.local` so it cannot re-enter dispatch), and the inline chain; an anonymous carrier accepts `.with` without a lookup (a blockless factory defines nothing but accessors).

**The measured trade, recorded in the suite rather than hidden by it:** the external corpus is invisible (five struct-bearing files from haml/hamlit/faraday/mail, ~11k lines — identical diagnostics and coverage before/after; a 170-site static census's chained reads all have non-struct receivers by the time the gate runs). Inside this repo the cost is real: the FACTORY-METHOD idiom (`def build = Pair.new(...); build.items…`) loses member folding — three regression guards from #293's arc documented it, and each was rewritten to keep BOTH halves discriminating (direct-construction forms keep the original widening property, cross-file included and verified gate-independent; the factory route's decline is pinned explicitly beside the direct form that still fires the typo). The loss is an FN; the bug was an FP-class wrong value. The richer follow-up (accept a callee whose body provably cannot return `self` — the slice-5 scan already refuses bare `self` in return position) is filed separately.

ADR-48 updated in the same commit: the falsified premise, the slice-2 paragraph rewritten around the whitelist with the measured cost, and the #595 pointer resolved. Discrimination: reverting only `fresh_receiver?` fails exactly the 7 new examples; the #591 slice-5 suite stays byte-green in both arms.

Closes #595.


zonuexe:
Landing gates, all green:

- **Corpus FP diff** (full 11-target set, base = post-#596 arms, `--no-baseline --no-cache`, re-collected after the fix round): **+0 / −0** — confirming the author's five-file measurement at full scale.
- **Critical review** (independent, two rounds): R1 REQUEST-CHANGES — the extraction was parameterized into divergence: the freshness consumer's lookup fell back to an own-class-only table, so a module-INCLUDED self-returning `with` was invisible to the read gate while the grant arm refused it — the exact class the unification exists to prevent. Fixed by deleting the parameter: one `Scope#user_def_through_ancestors` walk serves both consumers (caches preserved as cache arguments, byte-faithful relocation), and the guard-swap discrimination now reds BOTH consumers from one edit. R2: **APPROVE** — with a reasoned deferral adjudication: the block-form-include blindness (`discovered_includes` unpopulated for factory-block bodies) is live but is index DATA, not a predicate fork — both consumers agree, the wrong-value surface strictly shrinks vs master, and the fail-closed alternative costs mass FNs on the legitimate block-form `.with` population; the discovery-side fix + the budget-exhaustion fail-open + the aliased-constant FN are all recorded on #599, and ADR-48 carries the carve-out sentence.
- What lands: `fresh_receiver?` trusts only materialisations (shared `StructMaterialization` predicate), killing the self-returning fluent-builder wrong-value family (#595 + the composed d3 shape) with the factory-method idiom's FN paid openly (three #293-arc guards rewritten to stay discriminating, verified gate-independent; recovery filed as #599).

Issue #316, opened 2026-08-08
Toplevel `def output` in one file captures RSpec's `output` matcher in unrelated spec files


zonuexe:
A project-wide toplevel `def` captures a same-named method that an included module provides at the call site. Minimal repro (2 files, same run):

```ruby
# helper.rb — an executable script, never loaded by the specs
require "json"
def output(obj)
  print JSON.dump(obj)
end
```

```ruby
# a_spec.rb
RSpec.describe "x" do
  it "prints" do
    expect { puts "ok" }.to output(/ok/).to_stdout_from_any_process
  end
end
```

Actual (v0.3.1 `c39e6675` and master `2c38c76b`, fresh cwd, `--no-cache`):

```
error: undefined method `to_stdout_from_any_process' for nil
```

At runtime `output` inside the example group is RSpec's matcher (an included module's method beats a private toplevel `Object` def in MRO), so the code is correct. The analyzer binds the call to `helper.rb`'s toplevel `def output` (returns `nil` via `print`) and fires on the chain.

The toplevel-`def` table registers the name project-wide without considering that the call site sits inside a block whose receiver context gains DSL methods via `include`. Any project helper whose name collides with a DSL matcher (`output`, `match`, `include`, …) hits this; 16 live diagnostics of this shape in dependabot-core's updater specs.

*Found via differential testing against the rigor-rs port (which is silent here); verified today on both the v0.3.1 tag and master.*

PR #603, opened 2026-09-01
Carry opaque_block_self through scope joins and roster every join field

### LINKED ISSUE #600: Scope#join drops opaque_block_self: any if inside a DSL block re-arms cross-file top-level def binding
Found by PR #596's review while auditing `Scope#join` for sibling omissions of the same bug class (#596's root was `struct_fold_safe_locals` omitted from the join constructor, falling back to the default): **`build_joined_scope` also omits `opaque_block_self`** — proven at scope level (`opaque before join: true / after join: false`). Every join inside a DSL block resets the #316/#319 decline gate (`bindable_top_level_def_for`, scope.rb:552), so a cross-file top-level def binds again after ANY `if` inside an RSpec-style block — the wrong-bind/FP direction the gate exists to prevent.

Pre-existing (not introduced by #596, which only fixed the struct set). Fix shape mirrors #596: carry the flag through the join (conservative direction: the join is opaque if EITHER side is), one-line + a scope_spec unit pinning the before/after, plus an integration fixture (an `if` inside a DSL block followed by a bare top-level-def-named call must still decline).

While in there: audit the REMAINING Scope constructor fields for join omissions systematically — two of this class have now been found one review apart; a spec that constructs two fully-populated scopes, joins them, and asserts every carried-through field non-default would end the class.

zonuexe:
Fixes #600: `build_joined_scope` dropped `opaque_block_self`, so any `if` inside a DSL block re-armed the #316/#319 cross-file top-level-def binding gate — the wrong-bind/FP direction. The carry is `||` where the neighbouring fold-safety set INTERSECTS, and the comment records why the two grants point opposite ways: fold-safety licenses a FOLD (keeping an arm-lacked grant could fold a stale constant → intersect), opacity licenses a DECLINE (keeping it costs only precision → disjoin).

The issue's systematic half ships too: all 21 constructor keywords audited against the join (no third omission — `opaque_block_self` was the only drop; 15 merge from both arms, 5 are receiver-only BY DESIGN incl. the shared-by-reference advisory tables, encoded as expected semantics so a future audit doesn't 'fix' them), and a class-ending spec: a reflected ROSTER requiring a recorded join rule for every constructor keyword (a new field with no row fails, forcing the semantics DECISION at add time), a no-field-left-at-default probe that NAMES the dropped field (against the pre-fix join: `got: [:opaque_block_self]`), and the non-vacuity inversion (an unpopulated scope must report every field). Integration: the two-file DSL-block repro with an `if` spliced ahead of the call — must-not-bind after the merge, with the genuine-top-level must-bind sibling proving the gate isn't wholesale.

Breakage proof: reverting only scope.rb reds exactly 3 scope + 1 integration examples. Diagnostic A/B over lib: byte-identical (the fix only declines more, and lib has little DSL-block cross-file-def code — the fixture is the evidence). Draft pending the landing gates.

Closes #600.


zonuexe:
Landing gates, all green:

- **Corpus FP diff** (full 11-target set, `--no-baseline --no-cache`): **+0 / −0**; the reviewer additionally verified the lib diagnostic A/B byte-identical on both arms.
- **Critical review** (independent, by the reviewer who found the omission in #596's audit): **APPROVE** — the original probe holds through the join in both argument orders; the `||`-vs-intersection direction verified in both halves (cannot manufacture opacity — probed; mixed-opacity arms proven structurally unreachable across all three `Scope#join` call sites, with `||` erring toward silence if one ever arose); the roster's reflection has real teeth (current constructor keywords, not a hand list; the no-default probe named exactly `[:opaque_block_self]` on the pre-fix arm); the receiver-only five adjudicated lossless-today by construction-site enumeration; the 3+1 breakage claim reproduced exactly. Three cosmetic nits recorded in the review (a duplicate self_type example, a comment wording, two fields whose join rule is pinned only via the shared bindings path — all FN-direction).
- CI: fully green on the final head (verified in-review); the master post-merge run will be watched to completion per the tightened rule.

Issue #597, opened 2026-09-01
Model an in-loop struct setter's per-iteration effect instead of gating wholesale — the mail ragel residual


zonuexe:
Successor to #589 (whose mechanism PR #596 corrected and fixed — see its closing comment): the mail ragel cluster's actual shape is a struct local taking member SETTERS inside a loop — 131 `address.<member> = chars(...)` sites inside `address_lists_parser.rb`'s ragel `while`. Today every such member read after (or between) iterations answers Dynamic via the `deferred_setter` gate, which is load-bearing as an all-or-nothing device (removing it serves stale construction values, proven).

The precise lever: model an in-loop setter's PER-ITERATION effect instead of gating wholesale. A member assigned on every path through the loop body from the same expression family can be summarized — after ≥1 iteration the member holds the join of {construction value, the setter RHS types}; with unknown iteration count the sound summary is that join, never the bare construction value. This is the loop-seam sibling of #560's added-value join (evidence the mutation adds must reach the carrier) applied to struct members rather than array elements, and it composes with #596 (the grant now survives the loop, so a summarized member type would actually be consulted).

FP hazards to design against: the #540-family stale-fold trap in reverse (joining too little re-creates always-falsey on `address.local == parsed_value` shapes); ragel RHSs are `chars(...)` → honestly Dynamic today, so the near-term corpus yield is the reads of members the loop does NOT set (already paid by #596) plus honest Dynamic-bearing joins for the set ones — the payoff grows as #524-family return inference improves `chars`.

Gate: the mail parser fixture family (member read after an in-loop setter answers the join, not stale construction value, not Dynamic-wholesale), deferred-setter must-still-decline for non-every-path writes, full-11 corpus with mail as the target.

zonuexe:
Implementation attempt withdrawn (PR #604, now the withdrawal record) — the review found four empirically confirmed unsound folds in the "loop-exit union IS the per-iteration summary" approach, each serving a value the program never holds: no loop-carried fixpoint for member state (one unrolling), setter writeback no-oping on union bindings (fires loop-free), break-path arms dropped, in-body reads folding the pre-loop binding. The full analysis, probes, and the four unstated side conditions the argument depended on are in ADR-56 WD2.8.

[...]

PR #604, opened 2026-09-02
Record the loop-setter withdrawal and correct two boundary rationales

zonuexe:
This PR now ships the WITHDRAWAL of its original approach, after review found four empirically confirmed unsound folds (each serving a value the program never holds — the worst class): B1 no loop-carried fixpoint for member state (one unrolling posed as a summary — the pre-slice-B bug recreated one level down), B2 `apply_setter_writeback` no-oping on union bindings while the new read consulted them (fires with ZERO loops), B3 break-path arms dropped, B4 in-body reads folding the pre-loop binding. The union read had no separable sound piece (B2 is loop-free), and the motivation was already gone: mail's `address` is rebound 27× where fold-safety requires exactly one write — across #589/#596/#597, three levers were aimed at that file and none could structurally reach it (#605 is the honest successor).

What lands: `lib` byte-identical to the fork point except one corrected comment — `for`'s boundary rationale (it is a boundary because a single unrolling is not a summary, not because of block scoping); the stale spec comment split into the two genuinely different decline reasons (a block's bindings never leave its scope; a loop's reach the continuation but only as one unrolling); and ADR-56 WD2.8 rewritten as the withdrawal record — the four probes, the four unstated side conditions the original argument depended on, the what-a-sound-version-needs list (member-map widening in the fixpoint, union-arm writeback that REPLACES, break-sink struct carriers, converged entry state on the check path), and the mail measurement that removes the motivation. All four unsound folds verified `Dynamic[top]` again; #596's untouched-member win verified surviving.

Refs #597.


zonuexe:
Landing gates:

- **Critical review** (independent, three rounds): R1 REQUEST-CHANGES — four empirically confirmed unsound folds in the original loop-setter approach (member state never entering the fixpoint = one unrolling posed as a summary; setter writeback no-oping on union bindings, loop-free; break-path arms dropped; in-body reads folding pre-loop bindings), with the precise diagnosis that the soundness argument depended on four unstated side conditions. The author's size-and-motivation assessment chose HONEST WITHDRAWAL over a slice-B-scale fixpoint change with no measured beneficiary (mail's `address` is rebound 27× — unreachable by any per-iteration summary; #605 owns that now). R3 on the withdrawal head: **APPROVE** — all five probes verified back to `Dynamic[top]`, lib byte-identical to the fork point except one corrected comment, #596's untouched-member win verified surviving, ADR-56 WD2.8 a faithful withdrawal record (four findings, four side conditions, the design bar for any future attempt), no fragment ships.
- **Corpus arms**: waived with reason — `lib` is comment-only vs the fork point; the earlier full-11 collection on the (now-reverted) behaviour head was +0/−0 in any case.
- CI: green on the final head. #597 carries the design bar and the re-establish-the-beneficiary-first guidance; the withdrawal analysis is quotable from WD2.8.

Issue #449, opened 2026-08-22
The ActiveSupport gem overlay omits Date#to_time, so date.to_time(:utc) is a false positive without the plugin


zonuexe:
A Rails project that does **not** activate `rigor-activesupport-core-ext` gets a false positive on correct code:

```
$ cat app/x.rb
require "date"
d = Date.today
t = d.to_time(:utc)

$ rigor check
app/x.rb:3:7: error: wrong number of arguments to `to_time' on Date (given 1, expected 0)
```

`date.to_time(:utc)` is ordinary ActiveSupport. rbs's bundled `stdlib/date` types `Date#to_time` as `() -> Time`, and nothing widens it on this path, so the arity check fires on working code. AGENTS.md puts that cost above the worst-case static reading, which makes this a defect rather than a known limitation.

The ADR-72 auto-overlay is the path that should have covered it. [`data/gem_overlay/activesupport/core_ext.rbs`](https://github.com/rigortype/rigor/blob/master/data/gem_overlay/activesupport/core_ext.rbs) declares a substantial `class Date` section — `current`, `yesterday`, `tomorrow`, `beginning_of_week` and the rest — and **no `to_time` at all**. Omitting it avoids the duplicate-declaration collision that #437 was about, but it leaves the arity gap wide open.

Surfaced while fixing #437: the plugin lane hit the collision, the overlay lane dodged it by declaring nothing, and only the plugin lane got looked at.

## The fix is probably the row #437 just landed

#437 resolved the same tension in the plugin's signatures with an **overload continuation**:

```rbs
%a{pure}
def to_time: (?Symbol form) -> Time | ...
```

`| ...` appends the ActiveSupport overload ahead of the stdlib one, so both arities resolve and no duplicate-definition error is raised. Two things to check before copying it here, because the overlay is a different loading path:

- **`| ...` requires the base declaration to exist.** In the plugin's case `Environment.for_project` guarantees it by merging `DEFAULT_LIBRARIES` (including `date`) into every environment that loads a plugin's `signature_paths`. Confirm the same holds for the gem-overlay path before relying on it — a `| ...` with no base declaration fails.
- **The two never co-occur** (`Environment::GEM_OVERLAY_PLUGIN_IDS` stands the overlay down whenever the plugin loads), so fixing the overlay cannot reintroduce #437's collision. Worth re-confirming rather than assuming.

## While you are there

[...]

PR #607, opened 2026-09-02
Narrow through safe-navigation chains and nil-free membership guards

### LINKED ISSUE #606: Three guard shapes the nil-narrowing does not model: safe-nav chain truthiness/comparison and include? membership
The single root behind all 7 false positives that #574's measurement round found `call.possible-nil-receiver` firing on correct mastodon code (adjudicated at file:line on the measurement branch; see #574's matrix comment): three guard shapes the narrowing does not model, so the guarded local keeps its nil arm inside the protected branch:

1. `if custom&.value.present?` — a safe-nav CHAIN result being truthy proves the safe-nav RECEIVER non-nil (`custom&.value` truthy ⇒ `custom` non-nil); mastodon `extended_description.rb:10`, `privacy_policy.rb:13` (×2 sites each).
2. `next unless status&.account_id == @account.id` — a safe-nav chain compared `==` to a non-nil value proves the receiver non-nil (nil&.x is nil, nil == non-nil is false); `fetch_featured_collection_service.rb:37`. Sibling: `return true if backup&.user.nil?` — the FALL-THROUGH proves `backup&.user` non-nil hence `backup` non-nil; `backup_worker.rb:23,27`.
3. `unless REACTABLE_TYPES.include?(object_type)` + early return — `include?` on a nil-free literal collection proves the argument non-nil on the continue path; redmine `reactions_controller.rb:59` (this one is also the single pure FP in #574's arm C).

Each is a bounded narrowing form with a clean witness condition; fixing them removes 7 corpus FPs at ZERO true-positive cost — strictly better than #574's gate change on the same diagnostics. Slice 1 (safe-nav-chain truthiness/comparison ⇒ receiver non-nil) covers 6 of the 7. Gate per form: must-narrow fixture + must-NOT-narrow sibling (e.g. `include?` on a collection that CONTAINS nil must not narrow), corpus FP diff expecting exactly the named removals.

zonuexe:
Implements #606, the zero-TP-cost root fix for the FPs #574's measurement round adjudicated. Two independent slices:

**Slice 1 — safe-nav chains** (`analyse_safe_nav_chain` + `_equality`, hooked LAST so every existing path keeps first claim): a `x&.y.suffix…` chain truthy proves `x` non-nil when each suffix either raises on nil (a method NilClass does not define — branch unreachable, narrowing vacuously sound) or is documented falsey on nil (`present?`); `x&.y == <provably-non-nil>` proves the same, with `provably_non_nil_type?` a whitelist that answers FALSE for Dynamic. **The soundness catch that reshaped the brief:** most of what NilClass DOES define is truthy (`nil.to_s` is "", `nil.nil?` is true), so "any suffix truthy" is unsound — `nil?` takes the OPPOSITE edge (falsey proves), `!=` flips, both pinned as must-not-narrow. And the 7th corpus site (`fetch_featured_collection_service.rb:37`, `== @account.id` with an OPAQUE operand) is deliberately NOT narrowed — an opaque operand may be nil and `nil == nil` reaches the branch; it keeps firing, with its own must-not-narrow spec naming the file.

**Slice 2 — membership** (separate commit + spec file so it drops independently): `collection.include?(x)` truthy narrows `x` non-nil only when `nil_free_collection?` SEES the elements (literal/Tuple, or Array/Set with an element type); Hash/Range excluded (their include? asks about keys/bounds); truthy edge only. On current master this removes nothing — it pre-empts the one pure FP in #574's arm C, so it is the enabler that turns option 1's +3 into +2.

Corpus expectation at landing: exactly −6 on mastodon (extended_description ×2, privacy_policy ×2, backup_worker ×2), 0 elsewhere. Both slices are narrowing-only on scoped edges — they can only REMOVE diagnostics; the FN exposure fails safe (`nilclass_method?` returns method-present ⇒ decline on any reflection error). Discrimination proven against base per slice (5+1 red / 2 red, negatives green on both arms). Measurement gotchas recorded in the specs (a Constant-folding fixture was measuring dead-code elision, not narrowing — readers now return ivars).

Draft pending the landing gates. Closes #606.


zonuexe:
Landing gates, all green:

- **Corpus FP diff** (full 11-target set, `--no-baseline --no-cache`): **+0 / −6**, and the six removals match the implementation's per-site prediction EXACTLY (mastodon extended_description ×2, privacy_policy ×2, backup_worker ×2; the seventh matrix site, `fetch_featured_collection_service.rb:37`, resists BY DESIGN — its `==` operand is opaque and an opaque operand may be nil, so narrowing there would be unsound; pinned as a must-not-narrow spec naming the file).
- **Critical review** (independent): **APPROVE** — the adversarial battery held everything in the model (multi-suffix composition `x&.y.to_s.empty?` correctly rejected; `frozen?`/`class`/`hash`/`respond_to?` decline via ancestry or the argument gate; the `nil?`-opposite-edge and `!=`-flip soundness catches verified; membership's nil-bearing/union/Hash exclusions verified; downstream folds structurally safe — the narrowing produces a Nominal, which the always-truthy rule exempts). The one demonstrated wrong-narrowing arm (NIT 1: `duplicable?`/`instance_values` absent from the ActiveSupport overlay, so the raise-gate mis-modeled `nil.duplicable? == true`) is closed in `4818df00` — both rows added to the gem overlay AND its plugin twin (the #449 drift lesson) and verified against the reviewer's own probe. NITs 3/4 (signature-driven falsey derivation replacing the `present?` hardcode; the user-`==` assumption on typed membership elements) stand as recorded follow-up candidates.
- Both slices are narrowing-only on scoped truthy/falsey edges — removal-only by construction; the FN exposure fails safe (reflection errors decline).

PR #612, opened 2026-09-02
Record probed-but-missing paths as run-cache absence dependencies

### LINKED ISSUE #577: Run-result cache records only successful reads: a file ADDED after a miss (db/schema.rb) is invisible to warm runs
Found by the PR #576 review (probe preserved in that review's notes): on a warm cache, ADDING a previously missing file is invisible — `IoBoundary#read_file` records only successful reads (lib/rigor/analysis/io_boundary.rb, the read-recording arm), so a run that probed for `db/schema.rb`, found nothing, and cached its result has no dependency edge to invalidate when the schema later appears. The rigor-activerecord reduced mode made this visible (schema-less cached run → add `db/schema.rb` → warm run still serves the reduced index and prints a now-false "schema not found" :info; `--no-cache` control immediately fires the column checks), but the review's attribution probe showed master exhibits the same staleness class — the root cause is infra, pre-existing, not #576.

This is the ADR-45 sound-dependency design's missing negative half: "the analysis depended on X being ABSENT" is a dependency too. Fix direction: record probed-but-missing paths in the run-result dependency set (an absence edge keyed by path), and invalidate when the path comes into existence. Bound the recording to deliberate existence probes (the IoBoundary surface), not every failed stat.

Gate: warm-run fixture — cold schema-less run, add schema, warm run must re-analyze (column checks fire) without `--no-cache`; the inverse (schema removed) presumably already invalidates via the recorded read, verify with a sibling control.

zonuexe:
Implements #577 — ADR-45's missing negative half. `IoBoundary#read_file` recorded a `FileEntry` only on a successful read, so a run that probed for `db/schema.rb`, found nothing, and cached its reduced-mode result had NO dependency edge for the schema's later appearance: warm runs served the stale reduced rigor-activerecord index plus a now-false 'not found' `:info` (verified in #576's review). Now a `File.binread` that raises `ENOENT`/`ENOTDIR` records an absence row — `FileEntry.absent(path:)`, the existing `:exists` comparator with value "false" — before re-raising. No validation code changed: `Descriptor#fresh?` already validates `:exists` rows by a single `File.exist?` (no stat tuple or digest to drift, so no thrash), and both the Runner's post-run dependency descriptor and `Plugin::Base#producer_dependency_descriptor` are built from the boundary's files, so the single recording point covers the run-result entry AND the plugin-producer entries.

Recording is BOUNDED to deliberate probes: only the not-there outcome (`EISDIR`/`EACCES` record nothing, as before — a directory at the schema path would thrash every run, pinned as a bound not a gap), only inside the trusted-read scope (policy check precedes the read), and within one boundary a content row is never replaced by an absence row (`||=`) while a successful read after a probe replaces the absence row (`=`) — the conservative direction for a mid-run mutation. `SCHEMA_VERSION` 7→8 (pre-8 entries carry no absence rows and would validate fresh across exactly this edit; every developer's `.rigor/cache` clears once on first run — #57 marker discipline).

Binding docs in the same commit: ADR-45 WD1 (absence is a dependency too); `docs/internal-spec/cache.md` (the `:exists`/`FileEntry.absent` paragraph, schema-version list corrected from the already-stale 'currently 6' to 8 with v7+v8 rows); `plugin-trust.md` (`#read_file` row, also corrected `:digest`→`:stat` per ADR-87); `plugin-cache-producers.md`.

Specs: the issue's three-fixture gate driving the real Runner against a real on-disk Store — cold schema-less → ADD `db/schema.rb` → warm run misses (hits:0/misses:1), unknown-column fires, disclosure retracts; no-churn control (hits:1, diagnostics byte-equal); schema REMOVED inverse (misses via the recorded read) — plus 6 IoBoundary examples (ENOENT/ENOTDIR absence rows, content-over-absence ordering both ways, out-of-scope and EISDIR bounds), 2 Descriptor, 1 Store, 1 producer-cache. Discrimination: reverting only the ENOENT arm reds exactly the four predicted examples; every control stays green.

Independent review: APPROVE (4 nits recorded: a scope-out follow-up — plugins that gate their probe with `File.file?` BEFORE `read_file` record no absence row, e.g. rigor-sidekiq's schedule scan; the unused `FileEntry::PRESENT` symmetry constant; a pre-existing `docs/compatibility.md` marker literal drift). Draft pending the landing gates (full-11 corpus arms + CI).

Closes #577.


zonuexe:
Landing gates, all green:

- **Corpus FP diff** (full 11-target set, base = fresh master arms, `--no-baseline --no-cache`): **+0 / −0** — as expected for a cache-dependency change (the arms run cacheless; the diagnostics cannot move).
- **Critical review** (independent): **APPROVE** — the recording bound (not-there outcome only, trusted-read scope only, content-over-absence ordering) verified; the three-fixture gate (add-after-miss invalidates warm; nothing-changed still hits; remove-after-hit still invalidates) verified; SCHEMA_VERSION 7→8 load-bearing. Four nits recorded: the scope-out follow-up where plugins gate probes with `File.file?` before `read_file` (no absence row — rigor-sidekiq's schedule scan), the unused `FileEntry::PRESENT` symmetry constant, and a pre-existing `docs/compatibility.md` marker literal drift — the first is worth its own issue.
- CI: the chained watcher merges only on a green final head and then verifies the master merge-commit run to its conclusion.

Issue #615, opened 2026-09-02
Array.new(n) types as elementless Array, so the block/loop seams close it over stores: wrong-precise after the #586 rule


zonuexe:
Found by the #586 branch review (probes preserved in its notes; identical on master and the branch — pre-existing): `Array.new(n)` and `Array.new(n, v)` with a non-literal size type as a BARE `Array` with no type args, which the ADR-56 block/loop seams read as an ELEMENTLESS seed — so the seam closes the carrier over the block's stores and serves a wrong-precise element type: `acc = Array.new(n); [1].each { acc.push(1) }; acc.first == 1` folds always-truthy (the runtime array holds `nil`s from `Array.new(n)`), and `acc = Array.new(n, "x"); ...push(1)...; acc.first.upcase` draws `undefined method 'upcase' for 1` on correct code. Two adjacent corners from the same review: a Union seed with a whole-variable Dynamic member (`out = unknown(out); out << x` in a block) loses the Dynamic arm (`Array[1 | 2]`); a `while` inside a block appending a body-local reads `Array[Dynamic[top]]` (imprecise, not wrong).

This is #531's adjacency (the constructor fold's dynamic-size answer) meeting #586's seam rule: the honest seed for `Array.new(n)` is `Array[nil]` (and `Array[V]` for the fill form / block form's return), never elementless. Fix direction: give the dynamic-size `Array.new` fold a real element type (per #531's design, but only the SEED role matters here — the Tuple-vs-Nominal read hazard #531 recorded stays), so the seams see a declared-like Nominal with an element arm and keep it under #586's rule. Must-not-fire: both probes above; must-still-succeed: literal-size `Array.new(3, 0)` keeps its tuple fold.

PR #616, opened 2026-09-02
Keep declared and pre-existing Dynamic seed arms through the block and loop joins

### LINKED ISSUE #586: A declared Array[untyped] parameter closes to a precise element type under block mutation, firing on correct code
Found during PR #581's review (probe in that review's notes): a parameter DECLARED `Array[untyped]` that is mutated through the ADR-56 block-capture path CLOSES to a precise element type — `#: (Array[untyped]) -> String` + `[1, 2].each { a.push(rand(9)) }` leaves `a` as `Array[Integer]`, and a subsequent `a.first.upcase` (correct Ruby — the declared untyped element licenses any value) fires `call.undefined-method` for Integer. This fires on CURRENT MASTER and on the #581 branch alike — it is a pre-existing over-eager floor-drop in the slice-C rederivation (`join_array_content`'s drop of Dynamic evidence when body stores contribute concrete classes), distinct from and adjacent to #581's B2 (which concerned a nested `Integer | untyped` union arm and is fixed there).

The declared `untyped` element is the signature's own statement that pre-existing contents are unconstrained; body stores add evidence about what the BODY added, never about what the array already held. The rederivation treating its body-complete view as world-complete is the same category error #581's ADR-56 WD2.5 note now documents for the straight-line seam.

Fix direction: when the pre-body element type carries a Dynamic/untyped arm that is DECLARATION-sourced (or more conservatively: any pre-body Dynamic arm at all), the slice-C rederivation must preserve it through the join instead of dropping it once concrete evidence appears. Interacts with #580's provenance mark (a floor-vs-declared distinction is exactly what the mark provides) — landing #580 first likely makes this a one-liner.

Gate: the repro must-not-fire; the ADR-56 precision fixtures (`block_path_stays_precise`, loop_body_fixpoint snapshot) must stay byte-green; corpus FP diff.

zonuexe:
Fixes #586: a parameter declared `Array[untyped]` (and its `Hash[untyped, untyped]` twin, and the `while`-loop seam as the baseline probe showed) mutated through the ADR-56 capture path CLOSED to a precise element type — `[1, 2].each { a.push(rand(9)) }` left `a` as `Array[Integer]` and `a.first.upcase` (correct Ruby) drew `call.undefined-method`. `ContentJoin.join_array_content`/`join_hash_content` dropped every top-level Dynamic seed arm once body stores contributed concrete classes. The join now drops NO seed arm (`drop_dynamic` removed): the drop had exactly one producer — `MutationWidening.widen_after_block` spelling an empty `[]` as `Array[untyped]` before the block seam read its seed — so the block seam now reads its seed from the scope captured immediately BEFORE `widen_after_block` (`pre_widen_scope`/`seed_scope:`), the same cure #581's B2 round applied to the loop seam via `pre_body`. Reading earlier means the seams meet a `non-empty-array[T]` Difference where they used to meet its widened base, so the join helpers read a Difference through to its base — which also repairs a B2 side-effect on the loop seam (`if xs.any?; while …; xs << 1; end` read `Array[String] | non-empty-array[String]` with the `1` missing; now `Array[1 | String]`).

The distinction the ADR headline depends on is by SEED SOURCE, not provenance: at the pre-widen scope an empty `[]` is still `Tuple[]` (zero elements) while a declared param is `Nominal[Array, [Dynamic]]` (one element) — so `block_path_stays_precise` keeps `Array[1 | 2 | 3]` and the `loop_body_fixpoint` snapshot keeps `Array[Integer]`, byte-green.

**Independent review: APPROVE** — and the review re-measured the 'accepted monotone trade' as FOUR master false positives fixed: a Dynamic seed (`a.push(x)` untyped then a block; an escaping yielder; a parenthesized receiver; a local seeded from a `-> Array[untyped]` sig) each closed to a wrong-precise `Array[2]`/`Array[1]` on master and drew `undefined method 'upcase'` on correct code — all quiet on the branch. The sharpest probe (a straight-line #581 floor nested inside a Nominal's element union feeding a block seam) reads identically on both arms and is load-bearing (`a.push(1); a.push("s"); [1].each { a.push(2) }; a[1].upcase` is correct Ruby, quiet only because the arm survives). Pre-existing follow-ups the review found on BOTH arms are filed separately (`Array.new(n)` typing as bare Array and closing over block stores → wrong-precise).

Specs: the issue repro verbatim as a sig-wired project fixture (`mutation_join_declared_sig/`: block form, while form, the Hash twin — must-not-fire), fresh-seed closes + GENUINE-UNDEFINED line-set siblings (must-fire), non-empty-seed appends (both seams), the `[x]` slot keeping its arm, plus 8 unit examples replacing the dead `drop_dynamic` block. Discrimination: python-swapping both lib files back reds exactly the 7 unit + 3 integration examples; the ADR-56 precision fixtures pass in BOTH arms. ADR-56 gains WD2.9 with two in-place corrections to WD2.5 text that became false. Draft pending the landing gates (full-11 corpus arms + CI).

Closes #586.


zonuexe:
Landing gates, all green:

- **Corpus FP diff** (full 11-target set, base = fresh master arms, `--no-baseline --no-cache`): **+0 / −0** — the widening only ADDS arms or reads a Difference to its base, so the exposure is opacity, never a new firing; and the review re-measured the 'accepted trade' as four master false positives fixed (Dynamic seeds that closed to wrong-precise `Array[2]`/`Array[1]` and drew `undefined method 'upcase'` on correct code).
- **Critical review** (independent): **APPROVE** — the issue repro quiet on a sig-wired project fixture (block AND while forms, the Hash twin); ADR-56's headline precision (`block_path_stays_precise` `Array[1|2|3]`, the `loop_body_fixpoint` `Array[Integer]` snapshot) byte-green on both arms, with the seed-source distinction (an empty `[]` is still `Tuple[]` at the pre-widen scope) verified as the mechanism; declared `Array[Integer]` byte-identical to master; the sharpest probe (a straight-line floor nested inside a Nominal's element union feeding a block seam) identical on both arms and load-bearing. Nits addressed: the ADR's `Array.new` / `Hash.new(default)` wording corrected; pre-existing residues filed as #615 (`Array.new(n)` typing elementless and closing over block stores — wrong-precise on both arms).
- The chained watcher merges only on a green final head and verifies the master merge-commit run to its conclusion.

Issue #618, opened 2026-09-02
A top-level def shadows a struct member inside the struct's own body: implicit-self dispatch consults top-level defs before self_type's own methods


zonuexe:
Found by the #590 branch review (identical on master and for the `class Line < Struct.new(:text)` spelling — pre-existing): a TOP-LEVEL def shadows a struct MEMBER inside the struct's own body. `def text; nil; end` at top level + `Line = Struct.new(:text) do def s; text.upcase; end end` types the in-body `text` as `nil` (the top-level def) and fires `call.undefined-method: undefined method 'upcase' for nil` on correct code — Ruby dispatches to the member accessor `Line#text`. Root: `ExpressionTyper#try_local_def_dispatch` (expression_typer.rb ~1115) consults `bindable_top_level_def_for` for every implicit-self call BEFORE the receiver class's own discovered methods, regardless of `self_type`. Fix direction: when `self_type` is a class with a discovered method (or a struct/data member accessor) of that name, the class's own method wins; the top-level binding applies only to genuinely unresolved names. Must-fire: the shape above types the member; must-still-bind: a top-level def called from a body whose class does NOT define it. Adjacent pre-existing gaps from the same review (record, low priority): `Const = Struct.new(:text) do … end.freeze` and `Const ||= Struct.new(…)` spellings are not routed through the meta-constant handler (still warn); a NON-meta constant-write block (`CONFIG = Foo.configure do … end`) is not entered as an opaque block while the local-assign form is (asymmetry: a cross-file top-level def binds inside the constant form).

Issue #319, opened 2026-08-08
`Class.new do … end` block body is analyzed in top-level scope, not as the anonymous class body


zonuexe:
The block body of `Class.new do … end` is analyzed as top-level code, not as the anonymous class's body. Minimal repro (runs cleanly under `ruby`):

```ruby
observer_class = Class.new do
  attr_reader :count
  def initialize(bucket); @bucket = bucket; end
end
o = observer_class.new([])
p o.count
```

Actual (v0.3.1 `c39e6675` and master `2c38c76b`), three diagnostics:

```
warning: unresolved toplevel call to `attr_reader`. …
error: wrong number of arguments to `new' on Object (given 1, expected 0)
error: undefined method `count' for Object
```

The `attr_reader` warning is the tell: the block body is treated as ordinary toplevel code, so the anonymous class never gains the block's methods, and its bare-`Object` typing produces the two follow-on errors on correct code. The block should be analyzed as a class-body scope attached to the new anonymous class, mirroring `class Foo … end`. 16 live diagnostics of this shape across concurrent-ruby specs and erb.

*Found via differential testing against the rigor-rs port; verified today on both the v0.3.1 tag and master.*

PR #619, opened 2026-09-02
Enter constant-assigned meta-class bodies as class bodies for the toplevel-call rule

### LINKED ISSUE #590: call.unresolved-toplevel fires on struct-factory block bodies' own member reads

zonuexe:
Fixes #590 — with the root cause narrower than the issue's framing: only the CONSTANT-WRITE meta-new form fired `call.unresolved-toplevel` on a body's own member reads (`Const = Struct.new(:text) do … end` and its `Data.define` / `Class.new` / `Module.new` twins — including `attr_reader` in a `Const = Class.new do` body, the very FP #319 retired for the local-assign spelling). The bare, local-assign and `class X < Struct.new` spellings were already silent. The constant-write body is now entered the way the #319 arm enters the local-assign one (`enter_meta_class_body`, a pure factor-out of `self_type_for_class_body([anon])`), so member reads, member setters, `define_method` and nested blocks inside the body resolve against the carrier instead of the toplevel.

**Adjudicated deviation, recorded here:** a genuinely unknown name inside such a body is now SILENT rather than still-warning. The rule's own contract (ADR-34 'class-body case stays closed'; `Scope#toplevel?` = `self_type.nil?`) excludes class bodies, a `Class.new`/`Struct.new` block IS class_eval'd, master is already silent on the same unknown name in every sibling spelling, and a member whitelist would fire on the inherited surface (`to_h`, `members`, included-module names) — new FPs on correct code. The DSL-block gate (#316/#319/#603) is untouched: an RSpec-style block's bare unknown call still warns and an `if` inside a DSL block still declines cross-file binding (verified on both arms).

**Independent review: APPROVE** — 25 probe fixtures (Data.define, keyword_init, |klass| block param, Class.new(StandardError) with super, Module.new, module-nested and class<<self-nested bodies, path writes, splat members) all silent/consistent with master's FPs removed and no new diagnostic; pre-existing residues filed as #618 (a top-level def shadows a struct member inside its own body — unsound on both arms; the .freeze/||= spellings; the non-meta constant-write block asymmetry). Discrimination: byte-swapping master's evaluator reds exactly 9 examples, master's indexer exactly 4 — matching the implementation's lists.

Draft pending the landing gates (full-11 corpus arms + CI). Closes #590.


zonuexe:
Landing gates:

- **Corpus FP diff** (full 11-target set, base = fresh master arms, `--no-baseline --no-cache`): **+1 / −2**, all in haml `lib/haml/parser.rb` `Line = Struct.new(:whitespace, :text, …) do … def strip!(from) self.text = text[from..-1] …`. The two removals are exactly the issue's FP family: `undefined method 'text=' / 'text' for singleton(Haml::Parser)` — the body's member reads resolved against the enclosing class. The one addition is the honest report that replaces them: haml's own `sig/haml/parser.rbs` declares `attr_accessor text: String`, and `String#[]` with a Range returns `String?`, so `self.text = text[from..-1]` is a genuine `argument-type-mismatch` against the project's declared signature (Steep reports the same site). Adjudicated: a wrong diagnostic pair replaced by a correct one, not a new false positive.
- **Critical review: APPROVE** (summarised in the description; residues filed as #618).
- The chained watcher merges only on a green final head after the rebase onto post-#616 master and verifies the master merge-commit run to its conclusion.

Issue #617, opened 2026-09-02
Block-return residues after #587: find/detect first-iteration nil, the cap-above-8 mutation family, compound-write tails as live FP, String#<< unwidened


zonuexe:
Residues surfaced by the #587 branch review (each verified IDENTICAL on master — pre-existing, not introduced), grouped so the next block-return lever starts here instead of rediscovering them:

1. **`find` / `detect` / `index` / `find_index` with a rebound-capture predicate answer `nil`**: `seen = 0; [1, 2].find do |e| seen += 1; seen == 2 end` → `nil` (runtime `2`); `index` → `nil` (runtime `1`); the one-liner `find { |e| (seen += 1) == 2 }` → `nil`. Mechanism: #587's (b) makes the per-position predicates non-Constant so `assemble_find_result` correctly declines, but the FALLBACK block-return path types the block from the entry scope (first-iteration pin → `Constant[false]`) — WD2.9's residue (1) reaching a TUPLE receiver whenever the per-element assembler declines. `select`/`reject`/`filter_map` on the same shape decline soundly post-#587.
2. **The per-element threading cap (8) leaves the content-mutation family in place above it**: `([[]] * 9).map do |a| a << 1; a end` → nine stale `[]` positions. #584's cliff comment promises `Dynamic[top]` above the cap — true for body-locals, not for a mutated PARAMETER whose entry binding is the literal. Cheapest honest fix: above the cap, floor a position whose tail `tail_depends_on_body_binding?` instead of typing it tail-only.
3. **Residue (2) is a LIVE FP, not just imprecision**: `r = [1, 2].map { total += 1 }` → `[1, 1]` fires `flow.always-truthy-condition` on both `r.first == 1` AND `r.last == 1` (the second is false at runtime). Root: `type_of_assignment_write` (expression_typer.rb ~315) types every compound write as its RHS from the entry scope. WD2.9 records it; it needs the fix (#587's (b) covers `total += 1; total` tails, not the bare compound-write-as-tail shape).
4. **Straight-line `String#<<` is not widened at all**: `s = +"ab"; s << "c"; s == "ab"` folds always-truthy on correct code, while the ADR-56 block-capture path DOES widen the same `s` to String. A MutationWidening gap (its `widen_for_mutator` responds to Tuple/HashShape/empty-witness Difference only) — the #560 family's String door.

[...]

PR #620, opened 2026-09-02
Thread block returns through captured mutation and widen per-element folds over rebound captures

### LINKED ISSUE #587: Block-return typing: content-mutation-blind gate and first-iteration pinning both keep wrong-precise values
From the PR #584 review (probes in its notes), two pre-existing wrong-precise families in block-return typing, recorded so they stop hiding behind "byte-identical to master":

1. **The relevance gate is blind to content mutation.** `outer = []; r = m.synchronize do outer.push(1); outer end` types `r` as `[]` (the runtime value is `[1]`, and the continuation itself correctly types `outer` as `Array[Dynamic[top]]` afterwards) — `push` is not a variable-write node, so #584's gate declines threading and keeps the entry-scope literal. Threading would actually FIX this (inside a threaded body, MutationWidening fires — verified). Fix direction: extend `tail_depends_on_body_binding?` to also fire when a non-tail statement invokes a content-adder (`MutationWidening::CONTENT_ADDERS`) on a name the tail reads. Cheap, monotone (converts a wrong-precise to the widened honest answer).

2. **First-iteration pinning under multi-iteration per-element folds.** `total = 0; r = [1,2].map do total += 1; total end` → both master and the #584 branch answer a wrong-precise per-element value (`[0, 0]` / `[1, 1]`; runtime `[1, 2]`): the per-element Tuple fold types every element from the same entry scope, and the ADR-56 fixpoint's widened bindings never feed the return-typing pass (`call_type` is computed before the fixpoint runs — statement_evaluator's eval_call pipeline). Fix direction is architectural: either feed the fixpoint's widened exit bindings back into a second return-typing pass, or have the per-element fold decline (answer the joined/widened element) when the body writes a captured outer local. The always-falsey hazard is live: `r.first == 1` folds on a wrong pin.

Gate for either: must-not-fire on the two probes above, ADR-56 fixtures byte-green, corpus FP diff.

zonuexe:
Implements #587 as two independently droppable commits.

**(a) 688a636f — the #584 gate is blind to content mutation.** `tail_depends_on_body_binding?` collected only variable-write nodes, so `outer = []; r = m.synchronize do outer.push(1); outer end` declined threading and answered the entry literal `[]` (runtime `[1]`). The gate now also fires when a non-tail statement invokes a content adder (`MutationWidening`'s canonical mutator tables) on a name the tail reads — threading then runs the body through the main pass's `StatementEvaluator`, so `widen_after_block` + the content join answer `Array[Dynamic[top] | Integer]`. The #584 jump-decline scan and per-element cap are untouched (all 42 #584 examples byte-green).

**(b) fd827dc8 — first-iteration pinning under multi-iteration per-element folds.** `total = 0; r = [1, 2].map do total += 1; total end` answered a wrong-precise `[1, 1]` (runtime `[1, 2]`) and `r.first == 1` folded always-truthy; the per-element fold typed every element from the same entry scope. A body that WRITES a captured outer local (the new `CapturedLocals` helper — the four consumers now share one definition) now types each element as what that local can be in ANY iteration (`[Integer, Integer]`), never pinning to the first; content-only mutation (`log << e`) is not a write, so the exact `[2, 4]`-style folds survive.

**Independent review: APPROVE, both commits** — 69 adversarial probes (nested-block adders on the same name thread soundly; different-name adders don't thread; member-chain receivers unchanged; `next`/`break` declines and the arity-cap cliff re-verified; `||=`, multi-assign, `_1`/`it`, destructuring, Range receivers, rescue bodies, conditional rebinds all sound). No third wrong-precise family found; every wrong-precise answer still reachable verified byte-identical on master and filed as #617 (find/detect first-iteration nil, the cap-above-8 mutation family, compound-write tails as a live FP, straight-line `String#<<` unwidened). Non-vacuity: master's engine under this spec reds exactly the 12 must-change examples; the (a)-only tree is green (droppability verified). CLI as a user: `Array[Dynamic[top] | Integer]`, `[Integer, Integer]`, and only the exact-fold control fires.

Draft pending the landing gates (full-11 corpus arms + the CPU delta the name-keyed scan may carry — measured at landing + CI). Closes #587.


zonuexe:
Landing gates:

- **Corpus FP diff** (full 11-target set, base = fresh master arms, `--no-baseline --no-cache`): **+0 / −0** — both commits change only what a block that mutates a captured local answers, and no corpus site of that shape carried a firing on either arm.
- **CPU delta** of the name-keyed scan (`--workers=4 --no-cache`, user+sys seconds, arms alternated rep by rep, base = master checkout): no measurable cost.

| target | rep | base | branch |
| --- | --- | --- | --- |
| textbringer | 1 | 4.31 | 4.28 |
| textbringer | 2 | 4.34 | 4.26 |
| textbringer | 3 | 4.33 | 4.32 |
| redmine | 1 | 21.13 | 19.98 |
| redmine | 2 | 20.32 | 20.17 |
| redmine | 3 | 20.33 | 20.51 |

- **Critical review: APPROVE, both commits** (summarised in the description; residues filed as #617).
- Rebase note: ADR-56 conflicted with #616 (both appended a WD section at the same anchor); resolved by keeping both and numbering this one WD2.10.
- The chained watcher merges only on a green final head after the rebase onto post-#619 master and verifies the master merge-commit run to its conclusion.

Issue #623, opened 2026-09-02
rigor-activerecord tableizes the full constant path: Blog::Post resolves blog_posts where Rails uses posts


zonuexe:
Found by the #583 delta review (byte-identical on master; harmless today because an unknown table yields an empty column set and the column check stands down — but it makes every namespaced model's column and alias assertions vacuous).

`ModelIndex#inflected_table_name` derives a namespaced model's table through `Inflector.tableize("Blog::Post")` → `blog_posts` and `Admin::User` → `admin_users`. Rails derives `posts` / `users`: `ActiveRecord::Base.table_name` uses `undecorated_table_name(model_name)`, which applies `demodulize` before tableizing, and only prepends a prefix when the enclosing module defines `table_name_prefix` (`module Blog; def self.table_name_prefix = "blog_"; end`).

Repro (fixture from the review, `scratchpad/review583/fxC2_ns_tables`): schema has `posts(headline)`; `module Blog; class Post < ApplicationRecord; alias_attribute :headline, :title; end; end`; `Blog::Post.where(headline: "x")` — the index looks up `blog_posts`, finds no columns, and the alias/column checks silently stand down; with a `blog_posts` table present in the schema they run, so the spec that made the alias assertion non-vacuous had to add the wrong table.

Fix direction: demodulize before tableizing, and honour a `table_name_prefix` / `table_name_suffix` declared on the enclosing module when the discoverer can see it (a literal `def self.table_name_prefix = "…"` or `mattr_accessor`); when the module declares one the discoverer cannot fold, mark the table name computed (no pin). Gate: `Blog::Post` resolves `posts`; with `def self.table_name_prefix = "blog_"` resolves `blog_posts`; a top-level `Post` unchanged; an STI child keeps its parent's table. Also check `Inflector.tableize` callers in the other plugins (`rg tableize plugins/*/lib`) for the same shape.

Issue #621, opened 2026-09-02
Rooted-key rendering + find/find('::…') retry survives on six sibling plugins' own indexes (#583's bug class)


zonuexe:
Follow-up from the #583 review (PR branch `derooted-model-fact-keys`). #583 fixed the rigor-activerecord `model_index` fact: the discoverer rendered a `class ::User` declaration as the rooted key `"::User"` and every consumer then papered over it with a `find(name) || find("::#{name}")` retry. The same bug class survives on the sibling plugins' OWN indexes (not the shared fact), each with the rooted `"::#{parts.join('::')}"` rendering plus the retry:

- rigor-actionmailer `analyzer.rb:56`
- rigor-activejob `analyzer.rb:38`
- rigor-sidekiq `analyzer.rb:47`, `sidekiq.rb:197`
- rigor-actioncable `analyzer.rb:59`
- rigor-activestorage `activestorage.rb:94-95`, `analyzer.rb:44-45`
- rigor-actionpack `controller_index` (`actionpack.rb:429`)

Direction, per #583: normalise at the producer (strip the leading `::` when the declaration is rendered; a rooted re-open of a plain-declared class must merge into the same entry, never clobber it — see `ModelDiscoverer#merge_redeclarations` for the union rules), then delete the consumer-side retry arms. Gate: one e2e spec per plugin with a `class ::Foo` fixture whose declared surface (a mailer's `default from:`, a job's `queue_as`, a worker's `sidekiq_options`, a channel's streams, an attachment, a controller's actions) is still typed when the class is declared rooted, plus the existing specs green.

Why it matters: a `class ::Foo` re-open in a later-sorting file silently drops the earlier declaration's metadata on the current last-wins indexes — a config-order-dependent false positive class, the exact shape #583's second commit fixed for models.

PR #624, opened 2026-09-02
Key rooted model declarations by their de-rooted name and merge reopens into one row

### LINKED ISSUE #583: Models declared 'class ::User' are keyed '::User' in the model_index fact — all three Hash consumers silently miss them
Found by the PR #582 review: a model declared as `class ::User` is keyed `"::User"` in the published `:model_index` fact (`constant_path_name`'s nil-parent branch in the discoverer), and while rigor-activerecord's own analyzer tolerates both spellings (`find(name) || find("::#{name}")`), none of the three Hash CONSUMERS — rigor-actionpack, rigor-factorybot, rigor-shoulda-matchers — has the fallback, so their lookups silently miss such models (a uniform false-negative: checks stand down, nothing fires wrongly).

Fix direction: normalize at the PRODUCER — publish the fact keyed by the de-rooted name (`"User"`), since every consumer anchors with the plain constant rendering; or, if the leading-`::` distinction is load-bearing somewhere, add the two-spelling fallback to the shared consumer read path (all three consumers + the producer's own analyzer read the same fact — a tiny shared helper would end the drift). Producer-side normalization is the smaller, contract-cleaner change; check nothing keys on the rooted spelling first.

Gate: an end-to-end spec in one consumer with a `class ::User` fixture (must-fire unknown-column through the fact); the other consumers' spec files green; `make check-plugins` clean.

zonuexe:
Fixes #583. The rigor-activerecord `model_index` fact rendered a `class ::User` declaration as the rooted key `"::User"`, and every consumer (the AR plugin's own index, rigor-shoulda-matchers, the incremental fact merge) papered over it with a `find(name) || find("::#{name}")` retry that still missed the de-rooted spelling in three Hash consumers — a rooted model was silently untyped, its `where` columns unchecked. Keys are now normalised at the producer (`declared_constant_name` strips the root; superclass names and the configured `model_base_classes:` are de-rooted with them, so `["::ApplicationRecord"]` still matches), `ModelIndex#find`/`model?` accept a rooted query once, and the five retry arms are deleted.

**Second commit — the review's blocker.** De-rooting made `class ::User` and a `class User` reopen collide on one key, and `ModelIndex.build`'s last-wins assignment then CLOBBERED the real declaration's metadata whenever the reopen sorted later in the glob (an `alias_attribute` lost → a false `unknown-column` on `User.where(login:)`; a pre-existing clobber for the plain-reopen spelling, widened by the de-rooting). `ModelDiscoverer#merge_redeclarations` now UNIONs same-name rows after the model-hood fixpoint (model-hood unchanged): associations/enums/aliases name-keyed last-wins exactly as the STI chain merge, scopes/validations/callbacks union, `superclass_name` first-non-nil, `table_name_computed` OR'd. Uniqueness is single-homed in the discoverer and documented on `build`.

**Third commit — the delta review's nit.** `self.table_name =` is an assignment, so the LATER declaration wins as at load time, and two disagreeing literals mark the name computed (no value pin; the columns are checked against the later name).

**Reviews:** round 1 REQUEST_CHANGES (the clobber above, fixed); delta review **APPROVE** — three-file reopens in all six glob orders, STI parent+child both reopened, namespaced vs root keys (`module Admin; class ::User` merges into ROOT `User`, never `Admin::User`), superclass-less reopen sorting first, warm-cache add/delete of a reopen file, duplicate DSL rows dedup — all sound; discrimination 10 of 15 new examples red on the first commit's lib. Pre-existing residues filed: #621 (the same rooted-key + retry pattern on six sibling plugins' own indexes), #623 (`Blog::Post` tableizes to `blog_posts`, Rails uses `posts`). Recorded, optional: a doubled `enum` on one column resolves last-by-glob (documented; a per-column value union would be strictly FP-safer).

Manifest 0.7.0 → 0.8.0 (a cached 0.7.0 entry can be missing the surface the merge restores). Draft pending the landing gates (corpus arms on the AR targets + CI).


zonuexe:
Landing gates:

- **Corpus FP diff** (redmine + mastodon, the two AR-typed targets; base = fresh master arms, `--no-baseline --no-cache`): **+11 / −0, every addition an `info`-severity `model-call` recognition trace and no warning-or-above change on either arm.** All eleven sites spell the receiver ROOTED — the exact spelling the fix admits: mastodon `::Follow.find_by(...)` / `::FollowRequest.find_by(...)` inside `ActivityPub::Activity::Follow` (where a bare `Follow` would name the activity class), `::Web::PushSubscription.where(...)`, `::Web::Setting.where(...)`; redmine `::Query.where(...)` inside `User`. On master these calls were invisible to the plugin (its finder retry did not reach the de-rooted key); now they are typed and column-checked, and none draws a diagnostic.
- **Critical review: APPROVE** after one REQUEST_CHANGES round (the reopen clobber, fixed in the second commit; the delta review's table-name nit, fixed in the third — details in the description).
- Rebase note: the spec file conflicted with #612's absence-dependency block (both append a `describe` at the file's end); resolved by keeping both.

Issue #626, opened 2026-09-02
rigor-railties: drop the ::Rails hard-code once #614's rooted resolution lands


zonuexe:
Follow-up that becomes actionable once BOTH #588 (branch `rails-surface-follow-ups`) and #614 (branch `rooted-constant-resolution`) have landed.

#588's second-round fix added `resolved_rails_constant_name` to `plugins/rigor-railties/lib/rigor/plugin/railties.rb`: for a `::Rails` receiver it hard-codes the top-level answer, because the engine then resolved `::Rails` lexically (#614) and a project's nested `MyApp::Rails` shadow captured it. #614 makes the engine answer `::Rails` from the top level on its own (`Reflection.resolve_constant_type(name, scope:, rooted:)`), so the plugin-side special case is redundant.

Do: delete the `::Rails` arm of `resolved_rails_constant_name` and let the reader gate go through the engine's answer for both spellings; the railties spec's `::Rails` examples (the nested-shadow one added on #588, and `spec/integration/plugins/railties_plugin_spec.rb:52` "types the root-qualified `::Rails` receiver identically") must stay green. While there, `rails_constant_receiver?` (`railties.rb:143-148`) re-implements a single-segment slice of `Source::ConstantPath.rooted?` — fold it onto the shared predicate.

Issue #622, opened 2026-09-02
Incremental: an unresolved constant receiver records no absence edge, so declaring the constant later leaves its readers stale


zonuexe:
Found while reviewing the #588 branch (railties reader gate), reproduced WITHOUT the plugin — an engine gap in the ADR-46 incremental dependency lane.

**Repro** (`--workers=0`, fresh `.rigor/cache` per sequence, warm runs via `--incremental`):

1. `lib/demo.rb` = `Rigor.dump_type(Rails.logger); Rails.logger.info("x")` with NO `module Rails` anywhere in the project → cold run: the receiver `Rails` is an unresolved constant, `Rails.logger` types `Dynamic[top]`, no diagnostic.
2. Add `lib/rails_ext.rb` = `module Rails; def self.logger; :sym; end; end`.
3. Warm run → still `Dynamic[top]`, exit 0. A full `--no-cache` run answers `:sym` and fires `call.undefined-method` on `.info`.

The consumer file is never re-analysed because an unresolved constant receiver records NO negative edge: `record_cross_file_method` / `read_missing(:method, …)` fire only when the receiver constant resolved to a scope and the method lookup missed; a missing constant short-circuits before either. The plugin-side twin of this (a contribution answering before the engine's dispatch reaches the negative edge) is being fixed on the #588 branch by recording `read_missing(:method, "Owner.name")` from `Reflection.discovered_method?`; this issue is the engine's own hole: a missing CONSTANT should record `read_missing(:constant, "Rails")` (and the `appeared_symbols` producer must emit a `constant:Rails` key for a new `module Rails`), so that declaring the constant later re-analyses every file that read it as missing.

`--verify-incremental` does not catch it: its even-indexed subset (`check_command.rb:178-200`) re-analyses the consumer file itself, so the stale path is never exercised — the discriminating spec goes through `Analysis::IncrementalSession` (baseline → edit → recheck closure). See #577 for the absence-row precedent (SCHEMA_VERSION bump if the key grammar grows).

Issue #614, opened 2026-09-02
A leading :: is resolved lexically: ::Rails inside module MyApp types as MyApp::Rails


zonuexe:
Diagnosed while fixing #588's review blocker (branch `rails-surface-follow-ups`, commit 43c46690): `Source::ConstantPath.qualified_name_or_nil` renders a PARENT-LESS `ConstantPathNode` (the `::Rails` spelling) as the bare name `"Rails"`, so `Scope#type_of` resolves `::Rails` LEXICALLY — through `Reflection.resolve_constant_type`'s `Module.nesting` walk — exactly as if it were written `Rails`. Ruby's semantics are the opposite: a leading `::` always names the TOP-LEVEL constant. Observable wrong type: `module MyApp; module Rails; def self.logger; :nested; end; end; def self.probe; ::Rails.logger; end; end` types `::Rails.logger` as `:nested` (the nested module's reader) — and downstream, `::Rails.logger.info("x")` drew `call.undefined-method` on correct code before the plugin hard-coded its own top-level answer for that one receiver.

Every other consumer of `::Foo` inherits the bug: any project that shadows a framework constant inside a namespace and reaches the framework one with `::` (the standard Ruby escape hatch) gets the shadow's type. Fix direction: make `qualified_name_or_nil` (or the resolver's entry) carry the rooted marker so `resolve_constant_type` skips the nesting walk and answers the top-level constant directly; then remove the plugin-side hard-code in rigor-railties' `resolved_rails_constant_name` (it says why it exists). Must-fire: the shape above types the framework constant; must-still-succeed: bare `Rails` inside `MyApp` still resolves to `MyApp::Rails` (lexical), `::` on a genuinely top-level-only constant unchanged. Interacts with #528's namespace synthesis (a synthesized `MyApp::Rails` prefix must not capture `::Rails` either). Corpus FP diff expected: removals only (shadowed-`::` sites) — worth checking mastodon's `::` usages.

PR #628, opened 2026-09-02
Rails-surface follow-ups: Duration-on-the-left, the project-defined Rails reader gate, and its incremental edge

### LINKED ISSUE #588: rigor-railties / activesupport-core-ext follow-ups from the PR #585 review (comment drift, crashing-expression typing, project-Rails shadow)
Three non-blocking findings from PR #585's review, recorded as one follow-up so they don't evaporate:

1. **Comment/code drift in the Duration arithmetic rule** (`plugins/rigor-activesupport-core-ext/lib/rigor/plugin/activesupport_core_ext.rb` ~148): the block comment promises "the ARGUMENT is checked before the receiver … must cost as little as possible", but both `scope&.type_of` lookups run unconditionally — add the `next nil if argument_kind.nil?` early exit the comment describes, or reword the comment.

2. **`duration <op> Time/Date` is typed `Duration` on expressions that raise at runtime**: `30.minutes + Time.now` dumps `ActiveSupport::Duration` while runtime `Duration#+` cannot coerce a Time (TypeError). Silent-lenient today (no FP reachable), but the rule claims a type for crashing code — tighten the numeric/duration receiver branch to also require `argument_kind` ∈ {duration, numeric}, with a must-decline spec.

3. **A project-defined top-level `Rails` (or a lexically nested `MyApp::Rails`) is hijacked by the syntactic railties gate** — `Rails.logger` retypes to `ActiveSupport::BroadcastLogger` over the project's own def. Verified zero diagnostics fire (BroadcastLogger is RBS-less → lenient), so this is precision-loss/lost-TPs in a self-contradictory configuration, not an FP — but a cheap hardening exists: decline the contribution when `scope.type_of(call_node.receiver)` resolves to a project-defined constant. Also add one sentence to the plugin README noting activation is `plugins:`-list only (no lockfile gate).

Adjacent pre-existing observation (no action decided): the sidekiq worker-index lookup misses calls written `::WelcomeEmailWorker.perform_async` (rooted name vs plain-keyed index) — same family as #583's producer-side question; resolve consistently with whatever #583 lands.

zonuexe:
Fixes #588 — the rigor-railties / rigor-activesupport-core-ext follow-ups recorded by PR #585's review, in separately droppable commits.

- **Duration on the left of a Time/Date** (`30.minutes + Time.now` and its siblings raise at runtime): rigor-activesupport-core-ext no longer types the expression as `ActiveSupport::Duration`.
- **A project-defined `Rails` reader** (`module Rails; def self.logger; …`): rigor-railties now leaves `Rails.logger` / `.cache` / `.configuration` / `.application` alone at sites where the project defines that very reader on its own top-level `Rails`, instead of retyping the project's answer away — and only there: a reopen that adds unrelated members (`mattr_accessor`, `module EngineExtensions`), a compact `class Rails::HealthController`, and a nested `MyApp::Rails` shadow reached through bare `Rails` all keep the framework types (round-1 review blockers, fixed).
- **The gate records its ADR-46 dependency** (round-2 review blocker): the reader gate reads the cross-file discovered-methods table through `Reflection.discovered_method?`, which answered before the engine's own dispatch could record the negative edge, so a warm `--incremental` run kept serving `BroadcastLogger` after the project added `def self.logger`. The facade now records `read_missing(:method, "Rails.logger")` on a miss whenever the recorder is active (homed in the facade so any plugin gating on a project definition gets the edge); `docs/internal-spec/reflection.md` states the MUST and the key grammar; the discriminating spec goes through `Analysis::IncrementalSession` (definer gains `def self.build` → the gated consumer is re-checked and the merged diagnostics equal a full run; an unrelated `def self.other` does NOT re-check it). Reproduced and re-verified through the real CLI, warm run = `--no-cache` oracle in both directions.
- Comment drift from #585 corrected (the header's `::Rails` claim, the fingerprint sentinel's rationale).

**Reviews:** three rounds — REQUEST_CHANGES (two gate-shape blockers), REQUEST_CHANGES (the incremental staleness above), **APPROVE** on the head (key grammar for both kinds verified; the new `require_relative` is off every boot-slim path — `--version`/`help` load 16 files, a warm slim `check` 40, `reflection.rb` absent from both; `DependencyRecorder.active?` is only ever true inside a recording session; the miss set is bounded). Recorded from the reviews: the `kind: :instance` answer across files covers only names the indexer publishes cross-file (the ADR-17 monkey-patch contract withholds plain instance `def`s), now stated in the doc; the top-level pseudo-class key (`method:<toplevel>#name` vs the engine's `toplevel:name`) has no caller today. Filed: #622 (the engine's own hole — a brand-new `module Rails` + reader declared later stays stale, an unresolved constant receiver records no edge), #614 (`::Rails` resolved lexically — fixed on its own branch), #626 (drop this plugin's `::Rails` hard-code once #614 lands). Manifest stays 0.2.0 (no new contribution kind; the engine-source SHA invalidates checkouts).

Draft pending the landing gates (corpus arms on the Rails targets + CI).


zonuexe:
Landing gates:

- **Corpus FP diff** (redmine + mastodon, the Rails targets; `--no-baseline --no-cache`): **+0 / −0 attributable to this branch.** Against the pre-#624 base the branch shows exactly #624's eleven `info`-severity `model-call` recognition rows (rooted `::Model` call sites), which this tree carries by rebase; nothing else moved. mastodon's survey config omits rigor-railties, so the reader gate is exercised by redmine only — no change there either (redmine defines no `Rails` reader of its own).
- **Critical review: APPROVE** after two REQUEST_CHANGES rounds (gate-shape blockers, then the incremental negative edge — both fixed and discriminated; details in the description). Nits taken: the instance-kind visibility contract stated in `docs/internal-spec/reflection.md`.

zonuexe:
Landing gates:

- **Corpus FP diff** (redmine + mastodon, the Rails targets; `--no-baseline --no-cache`): **+0 / −0 attributable to this branch.** Against the pre-#624 base the branch shows exactly #624's eleven `info`-severity `model-call` recognition rows (rooted `::Model` call sites), which this tree carries by rebase; nothing else moved. mastodon's survey config omits rigor-railties, so the reader gate is exercised by redmine only — no change there either (redmine defines no `Rails` reader of its own).
- **Critical review: APPROVE** after two REQUEST_CHANGES rounds (gate-shape blockers, then the incremental negative edge — both fixed and discriminated; details in the description). Nits taken: the instance-kind visibility contract stated in `docs/internal-spec/reflection.md`.
- Identity note: three commits carried a non-public author address after the interrupted workflow; they were rewritten to the public one and the branch force-pushed before landing (the PR never merged with them).

Issue #634, opened 2026-09-02
Class.new(Struct.new(:text)) reports argument-type-mismatch: the struct-factory carrier is not accepted as a Class


zonuexe:
Found in passing by the #618 review (byte-identical on master): `Anon = Class.new(Struct.new(:text)) do … end` — valid, common Ruby — draws

```
call.argument-type-mismatch: argument type mismatch at parameter 'superclass' of 'new' on Class: expected Class, got Struct.new(:text)
```

The `Struct.new(:text)` carrier (Rigor's struct-factory type, `Type::StructClass`) is not accepted where RBS's `Class.new(?Class superclass)` expects a `Class`. A `Struct.new(...)` result IS a Class at runtime (so is `Data.define(...)`, and `Class.new(...)` itself); acceptance must treat the struct/data factory carriers as `Class` subtypes for parameter matching. Must-not-fire: `Class.new(Struct.new(:a))`, `Class.new(Data.define(:a))`, `Class.new(Class.new)`; must-still-fire: `Class.new("x")`, `Class.new(Struct.new(:a).new)`; and the resulting anonymous class should keep the struct members for in-body dispatch (#618's item 4).

Issue #633, opened 2026-09-02
Own-method-over-toplevel-def veto residues: inherited attr_*/define_method, RBS methods on pre-Object ancestors, singleton ancestor walk


zonuexe:
Residues recorded by the #618 review (branch `struct-member-over-toplevel-def`, APPROVE); all byte-identical on master — the veto that lets a class's own method beat a same-named top-level `def` does not reach these sources, so the top-level `def` still binds ahead of what Ruby dispatches to:

1. **Inherited non-`def` discovered methods.** Superclass `attr_accessor :stext` (with `@stext = "s"` in `initialize`) or an included module's `attr_reader :mtext`, plus a top-level `def stext; nil; end` elsewhere: inside the subclass, `stext.upcase` → `nil` + `undefined method 'upcase' for nil`. Cause: `Scope#user_def_through_ancestors` (scope.rb ~666) reads only `discovered_def_nodes` via `user_def_for`, while `attr_*` / `define_method` / `alias` live only in `discovered_methods`; `instance_self_answers?` tests `discovered_method?` for the OWN class only. Fix: run the same ancestor BFS with `discovered_method?(ancestor, name, :instance)`.
2. **RBS-declared method on a non-Object ancestor.** `class MyErr < StandardError; def m; message.upcase; end; end` + top-level `def message; nil; end` → `nil` + FP (Ruby: `Exception#message`); `class MyList < Array; first`; `include Comparable; clamp(1, 2)`; a project `sig/` declaring the method on the superclass / included module. The own-class-only restriction exists so that `Object`/`Kernel` names do not retract the #316/#319 tier — but a top-level `def` IS `Object#name`, so any owner that precedes `::Object` in the MRO wins at runtime. Sound generalisation: veto when `definition.defined_in` appears before `::Object` in `Reflection.instance_definition(class_name).ancestors.ancestors` (the existing `Widget` + `inspect` discriminator, `defined_in ::Object`, still fires).
3. **Singleton twins have no ancestor walk.** `class C < B` with `def self.sstext` on `B`: `Y = sstext.upcase` in `C`'s body / `def self.m; sstext.upcase` → `nil` + FP (Ruby: `B.sstext`). Fix: walk `superclass_of` with `discovered_method?(…, :singleton)` / `singleton_def_for`.
[...]

PR #636, opened 2026-09-02
Let a class's own method beat a same-named top-level def

### LINKED ISSUE #618: A top-level def shadows a struct member inside the struct's own body: implicit-self dispatch consults top-level defs before self_type's own methods

zonuexe:
Fixes #618. `ExpressionTyper#try_local_def_dispatch` bound the project-wide top-level `def` table for every implicit-self call ahead of anything the enclosing `self` could answer — inverting Ruby's MRO, where a top-level `def` is a private method on `Object`, the LAST link. So `def text; nil; end` at top level plus `Line = Struct.new(:text) do def s; text.upcase; end end` typed the in-body `text` as `nil` and fired `undefined method 'upcase' for nil` on correct code.

The candidate lookup stays first (it is a hash probe and owns the ADR-46 cross-file dependency edge) and a veto follows it: when `self_type` is a class (or its singleton) that answers the name itself — a `Struct.new` / `Data.define` member, an own discovered method (`def` / `attr_*` / `define_method` / alias), a `def` reached through the ADR-24 ancestor walk, or an RBS instance method whose `defined_in` IS the class — the class's own method wins; the top-level binding applies only to genuinely unresolved names, and a nil `self_type` (toplevel, unmodelled block self) is untouched, so the #316/#319 DSL-block tier is unchanged. The own-class-only RBS restriction is load-bearing (an inherited-declaration test would match every `Object`/`Kernel`/`Enumerable` name and retract that tier), pinned by a spec; the sound generalisation (owners that precede `::Object` in the MRO) is recorded on #633. Composes with #619: `Const = Struct.new(:text) do … end` bodies now resolve their members (two must-fire examples red without the veto even with #619 present).

**Review: APPROVE** — 46 `dump_type` sites across four fixture projects; every HEAD-vs-master delta is wrong-precise → correct or wrong-precise → `Dynamic[top]` matching the no-shadow control; reopened RBS classes, module `self` (`module_function`, `extend self`), blocks inside instance methods and class bodies, `prepend`, `alias`, project `sig/`, private methods, arity mismatches, and the #319 collocated-DSL must-still-bind all verified. Discrimination: 10 of 15 examples red on the master lib; `make check` / `make check-plugins` provably unaffected (591 files, zero top-level `def`s in `lib/` and `plugins/`). Residues filed: #633 (inherited `attr_*`/`define_method`, RBS on pre-Object ancestors, singleton ancestor walk), #634 (`Class.new(Struct.new(:text))` argument-type-mismatch, found in passing).

Corpus arms: +0 / −0 attributable (no corpus site has the shadow shape). Draft pending CI.


zonuexe:
Landing gates:

- **Corpus FP diff** (full 11-target set, base = fresh master arms, `--no-baseline --no-cache`): **+0 / −0 attributable to this change.** The branch's arms differ from the base only by #619's haml rows (−2 FPs / +1 honest `String?` report at `lib/haml/parser.rb:216`), which this tree includes by rebase; no corpus site has a top-level `def` shadowing a class's own member, so the veto never fires there. The self-check gates are unaffected by construction: `lib/**` and `plugins/*/lib` carry zero top-level `def`s.
- **Critical review: APPROVE** (summarised in the description; residues filed as #633 / #634; the fragment's over-claim corrected).

Issue #640, opened 2026-09-02
Run-result cache ignores producer watch globs: a worker added inside a discovery root outside paths: keeps serving the old diagnostics


zonuexe:
Found by the #613 review (byte-identical on the parent; not touched by that PR). The run-result cache entry (`analysis.run-diagnostics`) is built by `Runner#build_run_dependency_descriptor` (`lib/rigor/analysis/runner.rb ~1084-1092`) from `boundary.cache_descriptor.files` ONLY — the plugin producers' `GlobEntry` rows (`watch:` globs, e.g. rigor-sidekiq's `app/workers/**/*.rb`) never reach it, although `Descriptor#fresh?` already knows how to validate `globs`.

Repro (`--workers=0`, `--cache-stats`, fresh `.rigor/cache`): `.rigor.yml` = `paths: [demo.rb]`, `plugins: [rigor-sidekiq]`; `demo.rb` = `NewWorker.perform_async(1)`; cold run → no `worker-call` (no worker yet); add `app/workers/new_worker.rb` (OUTSIDE `paths:`) → warm run: `analysis.run-diagnostics` **1 hit / 0 misses**, served with no diagnostics, while `plugin.sidekiq.worker_index` itself recomputed (`0 hits, 1 miss`); `--no-cache` ground truth is `worker-call@1`. The same mechanism made #613's parent recompute `scheduled_workers` yet serve stale diagnostics.

Fix: append the producers' glob rows (and any other descriptor kinds they carry) to the run-result descriptor — compose the boundary descriptor with each producer's `cache_descriptor` at `build_run_dependency_descriptor`, or have the run entry depend on the producers' cache KEYS (a producer miss ⇒ run miss). Gate: the fixture above through the real CLI (add-inside-root invalidates the run entry; nothing-changed hits), plus the #613 three-fixture pattern for a plugin whose discovery root is outside `paths:`.

Issue #630, opened 2026-09-02
rigor-dry-types / dry-schema / dry-validation / graphql read project files with File.read, bypassing the IoBoundary: no cache dependency of theirs is ever recorded


zonuexe:
Found by the #613 audit of every `File.exist?` / `File.file?` probe in `plugins/*/lib`: `rigor-dry-types` (`dry_types.rb:107/109/125/127`), `rigor-dry-schema` (`:169/171`), `rigor-dry-validation` (`:142/144`) and `rigor-graphql` (`:83/85`) hold NO `IoBoundary` at all and read project files with `File.read` inside a walk over `Dir.glob` output. So none of their inputs record a dependency — neither the #577 absence row nor the ordinary `:stat` row a boundary `read_file` records — and the run-result cache (ADR-45) and `--incremental` (ADR-46) cannot see an edit, an addition, or a deletion of the files these plugins derive their contributions from. That is a wider staleness class than #613's probe gap: a warm run keeps serving the previous run's dry/graphql-derived types after the schema or type definitions change, until `--no-cache`.

Fix: give the four plugins a boundary (`Plugin::Base#io_boundary` is already the lazy accessor; `read_file` records the read; `file?`/`directory?` from #613 record the probes) and route every `File.read` / `File.file?` / `Dir.glob` result they consume through it; the entry-classification probes inside the glob walk stay bare (the entry exists by enumeration) but the READ must go through the boundary. Gate per plugin: cold → edit the definition file → warm run reflects the edit; nothing-changed run hits; delete → warm run drops the contribution. Also a self-check rule candidate (the #601-style umbrella for "probes and reads that bypass the boundary"): `rg 'File\.read\b' plugins/*/lib` should have zero hits outside the boundary implementation.

zonuexe:
Started on this and stopped before implementing, because **two of the issue's premises did not survive checking**. Recording what I verified and what I could not, so the next attempt starts from evidence rather than from the grep this was filed on.

## 1. `rigor-dry-types` does not belong on the list

[...]

Issue #629, opened 2026-09-02
rigor-actionpack missing-template stays reported from the run cache after the template appears (locate_template probes with File.file?, no boundary)


zonuexe:
Found by the #613 implementation (branch `absence-edge-existence-probes`), reproduced through the real CLI, NOT fixed there — it is a live false positive on correct code under the run-result cache.

`plugins/rigor-actionpack/lib/rigor/plugin/actionpack/analyzer.rb:474` (`locate_template`) probes each candidate template path with `File.file?`, and the negative answer IS the `plugin.actionpack.missing-template` diagnostic. The `Analyzer` holds no `IoBoundary` (it is constructed per run with the view roots only), so the probe records no absence row (#577 / #613) and the cached result has no edge back to the template's later appearance:

1. cold run: `render :show` in `PostsController` → `plugin.actionpack.missing-template` (correct — no template yet)
2. create `app/views/posts/show.html.erb`
3. next run: served from the run-result cache (`analysis.run-diagnostics: 1 hit`) and STILL reports the template missing — a false positive that persists until `--no-cache`.

`abstract_base_controller?` (`analyzer.rb:398` / `:401`) is the same shape at lower stakes.

Two decisions this needs before the mechanical fix (route the probes through `IoBoundary#file?` / `#directory?` from #613):
- give the `Analyzer` a boundary (thread the plugin's `io_boundary` through its constructor, or have the plugin resolve templates and hand the analyzer a resolved map);
- cardinality: `locate_template` tries ~9 extensions × N view roots per `render` call, so a naive per-probe row adds thousands of `File.exist?` validations to every warm run on a mastodon-scale project. Options: record ONE directory-level row per `app/views/<controller>` directory (a listing-fingerprint row, so any file added or removed under it invalidates) instead of a row per candidate path; or record only the FIRST miss per render call. The directory-listing row is the precise one and keeps the per-run validation count at the number of controller view directories.

Gate: the #577 three-fixture pattern through the real CLI (add-after-miss invalidates; nothing-changed hits; remove-after-hit invalidates) on the missing-template rule, plus a warm-run validation count assertion so the cardinality decision is pinned.

PR #641, opened 2026-09-02
Record plugin existence probes as run-cache dependencies

### LINKED ISSUE #613: Plugins that gate a probe with File.exist? before read_file record no absence dependency — the #577 staleness survives there
Follow-up from PR #612's review (#577, absence-edge cache dependencies): the absence row is recorded ONLY when `IoBoundary#read_file` itself hits ENOENT/ENOTDIR. Several plugins gate their probe with `File.file?` / `File.exist?` BEFORE calling `io_boundary.read_file` (e.g. `plugins/rigor-sidekiq/lib/rigor/plugin/sidekiq/schedule_scan.rb:71` — `return nil unless File.file?(absolute)` for `config/sidekiq.yml`; the review named rigor-activestorage's probe as the same shape), so a missing file records NOTHING and the #577 staleness class survives there: a warm run keeps serving the no-config result after the file appears.

Fix direction: route existence probes through the boundary — either a boundary-level `exist?`/`file?` that records the absence row itself (widens the hand-sig'd IoBoundary surface + drift snapshot + docs, deliberately not done in #612), or a lint/self-check rule that flags `File.exist?`/`File.file?` on a boundary-managed path inside plugins (the #601-style umbrella for "probes that bypass the boundary"). The first is the honest fix; audit every plugin's probe sites (`rg 'File\.(exist|file)\?' plugins/*/lib`) and convert them in one PR, with the #577 three-fixture gate reused per plugin (add-after-miss invalidates; nothing-changed hits; remove-after-hit invalidates).

Also from the same review (trivial, fold into this PR): `FileEntry::PRESENT` has no production reader (drop or leave); `docs/compatibility.md:76,:103` still show the marker literal `<version>.4.2` (now `<version>.8.2`).

zonuexe:
Fixes #613. PR #612 (#577) records a run-cache absence row when `IoBoundary#read_file` hits ENOENT — but the prevailing plugin shape gates the read on a bare `File.file?` / `File.directory?` first, so a missing `config/sidekiq.yml` or an absent `app/models` root recorded nothing and a warm run kept serving the no-config result after the file appeared.

**The probe is the read.** `IoBoundary#file?` / `#directory?` are that probe with the row attached: `FileEntry.present` on a hit (so a probed-but-never-read path — a globbed discovery root, an empty `app/models` — still has an edge back to the filesystem), `FileEntry.absent` on nothing-there, and nothing when something exists but is the wrong kind (#577's `EISDIR` bound in probe form). They answer truthfully everywhere and never raise, because they replace bare `File` predicates in plugin code — a raising predicate would change what the plugin does, not only what it records; outside the trusted-read scope they answer and record nothing (recording, not the answer, is what the policy gates — `docs/internal-spec/plugin-trust.md`). `FileEntry.present(path:)` gives `FileEntry::PRESENT` its production reader.

**19 probe sites in 12 plugins converted** (sidekiq's `config/sidekiq.yml`, activejob's `config/recurring.yml` and `config/application.rb`, sorbet's source roots + `sorbet/rbi`, factorybot's definition paths, and the `directory?` discovery-root gates of activestorage, actionpack's controller discoverer, actioncable, actionmailer, activerecord, pundit, the sidekiq worker scan, rails-i18n, rails-routes). Left alone with reasons in the commit: entry classification inside a `Dir.glob` walk (dry-types/schema/validation, graphql — which read with `File.read` and hold no boundary at all: #630), the actionpack `Analyzer`'s template probes (no boundary, and a cardinality decision: #629), rbs-inline's analyzed-file check, sorbet's path canonicalisation, plugin-load paths outside any project scope. Surface wiring in the same commit: `sig/rigor/plugin/io_boundary.rbs`, the API-drift snapshot, `plugin-trust.md`, `cache.md`, `plugin-cache-producers.md`, ADR-45 § WD1b. `SCHEMA_VERSION` deliberately not bumped: no row kind, comparator or value grammar changes, and version 8 (#612) has never shipped (v0.3.6 carries 6). The `docs/compatibility.md` marker literal corrected to `<version>.8.2`.

**Review: APPROVE** — the CLI three-fixture sequence reproduced on both arms (parent: config added → 1 hit, stale; branch: 0 hits / 1 miss, re-analysed; nothing-changed hits; removal misses), through the boot-slim `RunCacheProbe` → `Descriptor#fresh?` path; out-of-scope semantics verified not a trust regression; ordering (`||=` for existence rows, `=` for content rows; disagreeing rows across boundaries make the entry never-fresh, never a wrong hit); all 16 discoverer receivers are live boundaries. Discrimination: reverting only the sidekiq site reds exactly the add-after-miss example; reverting the boundary + descriptor reds exactly the 8 new unit examples. Pre-existing staleness found while probing, filed: #640 (the run-result entry ignores producer `watch:` globs), #629 (stale `missing-template`). Nits taken: ADR-45 WD1b names the actionpack exclusion and the `:exists` kind bound.

Corpus arms: +0 / −0 (a `--no-cache` arm cannot differ). Draft pending CI.


zonuexe:
Landing gates:

- **Corpus FP diff** (full 11-target set, base = master arms at the branch's parent, `--no-baseline --no-cache`): **+0 / −0.** Existence probes only change what a WARM run may serve; a `--no-cache` arm cannot differ, and it does not — the branch is byte-identical to its parent on every target.
- **Critical review: APPROVE** (summarised in the description; pre-existing staleness filed as #640 / #629; nits taken in ADR-45 WD1b).

Issue #639, opened 2026-09-02
--verify-incremental followed by deleting a declaring file leaves the next --incremental run stale


zonuexe:
Found by the #614 review, pre-existing (reproduces with a bare, un-rooted constant on master and the branch alike):

1. cold run with `lib/top.rb` (`class Foo; end`) and a consumer reading `Foo`
2. `rigor check --incremental` (warm, correct)
3. `rigor check --verify-incremental` → OK
4. `rm lib/top.rb`
5. `rigor check --incremental` → still types `Foo` (stale); a `--no-cache` run answers `Dynamic[top]` / the honest unresolved answer.

The same sequence WITHOUT step 3 (`--verify-incremental`) invalidates correctly on the delete, so the verify run leaves the incremental manifest in a state where the later deletion of the declaring file is not seen — likely the verify run's even-indexed re-analysis subset (`check_command.rb ~178-200`) writes back a snapshot / manifest whose file set or dependency rows no longer carry the consumer's edge to `lib/top.rb`. Fix direction: `--verify-incremental` must not persist anything that a subsequent `--incremental` run reads (it is a check, not a producer), or it must persist the FULL dependency set it verified. Gate: the five-step CLI sequence through `Analysis::IncrementalSession` (delete-after-verify invalidates), plus the existing verify examples.

zonuexe:
**This issue's stated root cause is wrong.** Measured twice during the [#644](https://github.com/rigortype/rigor/issues/644) work — once by its implementer, then independently by its adversarial reviewer — both on master and on the #644 branch.

The `--verify-incremental` step is **not** what causes the staleness. With `class Foo` in `lib/top.rb` and a reader `Rigor.dump_type(Foo)` + `Foo.zzz`:

```
cold  →  warm  →  rm lib/top.rb  →  --incremental
```

still prints `dump_type: singleton(Foo)` while the `--no-cache` oracle says `Dynamic[top]` — **with or without** the `--verify-incremental` step. The issue asserts that the sequence without step 3 invalidates correctly; it does not. It is also stale in **editor mode** on master: a buffer removing `class Foo` reports `re-analysed 1 file(s), 1 served from cache` and still answers `singleton(Foo)`.

## Actual mechanism

[...]

Issue #638, opened 2026-09-02
class ::Foo declared inside module MyApp is indexed as MyApp::Foo: rooted declarations take the lexical prefix


zonuexe:
Found by the #614 review (byte-identical on master). A rooted DECLARATION inside a namespace is indexed under the lexical prefix: `module MyApp; class ::Foo; def top; 1; end; end; end` registers `MyApp::Foo` (merged with any real `MyApp::Foo` shadow) instead of the top-level `Foo` that Ruby actually defines/reopens. Observable: with a real `module MyApp; class Foo; def nested; :n; end; end; end` present, `Foo.new.top` (top-level) types `Dynamic[top]`, while `MyApp::Foo.new.top` types `1` and `MyApp::Foo.new.nested` `:n` — the two classes are conflated under the wrong key. (rigor-activerecord's `model_index` fact fixed the same shape for models in #624; the engine's own `discovered_classes` keys were never de-rooted.)

Fix: the declaration renderer (`ScopeIndexer`, the `class`/`module` entry path) must key a rooted constant path at the top level — `::Foo` → `"Foo"`, `::A::B` → `"A::B"` — regardless of the lexical nesting, and enter its body with `self_type = Foo`; nested references inside such a body follow Ruby (lexical scope is still `MyApp` for constant lookup — `Module.nesting` inside `class ::Foo` within `module MyApp` is `[Foo, MyApp]`). Must-fire: the repro keys `Foo` at the top level and `MyApp::Foo` separately; must-still-succeed: `class Foo` inside `MyApp` stays `MyApp::Foo`; the ADR-46 `appeared_symbols` key for the declaration must match (a consumer that read a missing top-level `Foo` is re-checked when `class ::Foo` appears inside a module — pairs with #622).

zonuexe:
Fixed on master. `Source::ConstantPath.declaration_prefix` re-anchors a rooted header, and `spec/integration/rooted_ancestor_name_spec.rb` plus `spec/integration/fixtures/rooted_declaration_header.rb` and `spec/rigor/source/constant_path_spec.rb` all name this issue.

Re-run against master with a discriminating fixture — `class ::Foo` AND a genuine `MyApp::Foo` declared in the same `module MyApp` body, each owning a different method, so a wrong index would cross them:

```
f638.rb:11:7: info: dump_type: :nested   # MyApp::Foo.new.nested_val
f638.rb:12:7: info: dump_type: :rooted   # ::Foo.new.rooted_val
```

[...]

Issue #637, opened 2026-09-02
class X < ::Base resolves the superclass lexically: a rooted superclass inside a shadowing namespace inherits the shadow


zonuexe:
Found by the #614 review (byte-identical on master; #614 fixed constant READS only). The superclass / ancestry path is a lexical consumer of the rooted marker that the read-side fix did not reach: `scope_indexer.rb ~2009` (`class X < ::Base`) and `~2037` (`Class.new(::Parent)`) de-root the superclass name through the lenient `ConstantPath.qualified_name`, and `Scope#compute_ancestor_class_name` (`scope.rb ~722`) / `Reflection.resolve_ancestor_name` (`reflection.rb ~254`) then resolve it lexically.

Repro: `class Base; def top_base; 1; end; end; module MyApp; class Base; def nested_base; :n; end; end; class X < ::Base; end; end` → `MyApp::X.new.nested_base` types `:n` (the shadow's method — wrong; Ruby's `X` inherits the TOP-LEVEL `Base`) and `MyApp::X.new.top_base` types `Dynamic[top]`.

Fix: carry `Source::ConstantPath.rooted?` (from #614) from the superclass expression into the recorded ancestor name so `compute_ancestor_class_name` / `resolve_ancestor_name` skip the nesting walk for a rooted superclass; same for `Class.new(::Parent)` and `Struct.new`-with-superclass spellings. Note the lenient `qualified_name` has ~30 consumers (`rg 'ConstantPath\.qualified_name\('`) — audit which are lexical resolvers (need the marker) versus key renderers (must stay un-rooted, #583). Must-fire: the repro types `top_base` → `1` and `nested_base` → `Dynamic`/undefined; must-still-succeed: bare `class X < Base` inside `MyApp` inherits `MyApp::Base`.

zonuexe:
Fixed on master by [#754](https://github.com/rigortype/rigor/pull/754) (landed as `#722` residue 1, which also names this issue). Closing with the repro re-run rather than on the PR title.

`Scope#ancestor_name_candidates` short-circuits a rooted ancestor name to a single candidate — `raw.delete_prefix("::")` — so `class X < ::Base` names `::Base` wherever it is written, and `Inference::ScopeIndexer.recorded_ancestor_name` preserves the leading `::` that marks it. `spec/integration/rooted_ancestor_name_spec.rb` pins it.

Re-run against master, with a shadowing `NS::Base` present:

```
[...]

Issue #627, opened 2026-09-02
Diagnostics inside the dead arm of a version guard (Gem::Version / RUBY_VERSION comparisons) are false positives on the target Ruby


zonuexe:
Surfaced by the #614 corpus arms (branch `rooted-constant-resolution`): once `::YAML` resolves to Psych instead of the lexical `Mail::YAML` shadow, `mail/lib/mail/yaml.rb:26` draws `call.wrong-arity: wrong number of arguments to 'safe_load' on YAML (given 2, expected 1)`.

The report is honest against the Ruby 4 Psych signature (`safe_load(String, ?permitted_classes: …)`), but the line is the dead arm of a version guard:

```ruby
if Gem::Version.new(Psych::VERSION) >= Gem::Version.new('3.1.0.pre1')
  ::YAML.safe_load(yaml, :permitted_classes => permitted_classes)
else
  ::YAML.safe_load(yaml, permitted_classes)   # Psych < 3.1 positional form — never runs on Ruby >= 3.1
end
```

Multi-version gems (mail, kramdown, concurrent-ruby, …) carry this shape routinely: `RUBY_VERSION`, `Gem::Version.new(X::VERSION)`, `defined?(Ractor)`, `RUBY_ENGINE` comparisons selecting an arm that is dead on the target Ruby. A diagnostic inside the dead arm is a false positive for the user running Rigor on that Ruby — "the program works" outranks the worst-case static reading.

Fix direction (FP-first, narrow): fold `Gem::Version.new(<literal>) <cmp> Gem::Version.new(<literal>)` and `RUBY_VERSION <cmp> <literal>` when BOTH sides are literals or version constants Rigor can read (`RUBY_VERSION`, `Psych::VERSION` via the bundled gem / RBS), and treat the dead arm as unreachable for diagnostics (ADR-47 clause reachability already carries the mechanism for `if false`-style arms). Do not attempt symbolic version reasoning; an unfoldable guard keeps both arms live. Must-fire: an arity error in the LIVE arm still reports; must-not-fire: the mail shape above; discrimination: swap the guard's constant so the other arm is live.

PR #642, opened 2026-09-02
Resolve a rooted constant reference from the top level

### LINKED ISSUE #614: A leading :: is resolved lexically: ::Rails inside module MyApp types as MyApp::Rails

zonuexe:
Fixes #614. `Source::ConstantPath.qualified_name_or_nil` renders a parent-less `ConstantPathNode` (the `::Rails` spelling) as the bare name — deliberately, because every discovered-constant table is keyed un-rooted — so the root marker was lost between parse and resolver and `Reflection.resolve_constant_type` walked `Module.nesting` for a reference whose entire purpose is to skip it. `module MyApp; module Rails; def self.logger; :nested; end; end; def self.probe; ::Rails.logger; end; end` typed `::Rails.logger` as `:nested`, and `::Rails.logger.info("x")` drew `call.undefined-method` on correct code (the #588 review blocker's shape).

The marker now travels out of band — no `"::Foo"` string ever enters a key: `ConstantPath.rooted?` (leftmost segment, so `::A::B` works) is passed as `rooted:` to `Reflection.resolve_constant_type`, which then answers `constant_type_at(name, scope)` directly, skipping the nesting walk AND the #354 ancestor step; no top-level owner ⇒ the unresolved answer (`Dynamic`), never the shadow. Consumers carrying it: `ExpressionTyper#type_of_constant_path`, `Narrowing`'s Regexp match-operand resolution, and (the review's nit) `is_a?`/`kind_of?`/`instance_of?` predicates, which decline rather than narrow to the shadow when no top-level class owns the name; `case`/`when` and `===` already answered top-level through the environment-level `class_pattern_certainty`. The shadowed-rescue collector's private `absolute_path?` now delegates to the shared predicate. Declaration-side renderers keep un-rooted keys. `sig/rigor/reflection.rbs` + the drift snapshot; `docs/internal-spec/inference-engine.md` gained step 0 of lexical constant lookup and records the reachability divergence.

**Review: APPROVE** — the rooted answer verified for classes, modules, RBS core classes (`::String`, `::File`), top-level aliases, inside `class << self`, blocks, and nested namespaces; the nil branch (`::Foo`, `::Foo.new`, `::Foo::BAR`) `Dynamic[top]` with zero diagnostics; multi-segment `::A::B` from the root and `::C::D` unresolved when the top-level `C` lacks `D`; #528's synthesized `MyApp::Rails` prefix does not capture `::Rails`; ADR-46 edges unchanged (constant reads record none on either arm; warm-run add/remove probes correct). Discrimination: 16 of 322 examples red on the parent lib. Pre-existing residues filed: #625 (`rigor unused` still resolves `::Foo` lexically), #637 (`class X < ::Base` inherits the shadow), #638 (`class ::Foo` inside a module is keyed `MyApp::Foo`), #639 (`--verify-incremental` then delete stays stale); #626 (drop rigor-railties' `::Rails` hard-code, now removable).

**Corpus arms (full 11):** +1 / −4 — removals at shadowed-`::` sites (kramdown's `::StringScanner` inside `Kramdown::Utils::StringScanner` ×3, concurrent-ruby's `::Hash` ivar write); the one addition is mail's `::YAML.safe_load(yaml, permitted_classes)` `wrong-arity`, honest against the Ruby 4 Psych signature once `::YAML` reaches Psych instead of `Mail::YAML` — the line is the dead arm of a `Psych::VERSION < 3.1` guard, filed as #627 (version-guard dead-arm pruning). Draft pending CI.


zonuexe:
Landing gates:

- **Corpus FP diff** (full 11-target set, base = fresh master arms, `--no-baseline --no-cache`): **+1 / −4.** The four removals are exactly the predicted shape — a `::Const` reference captured by a nested shadow: kramdown's `::StringScanner` inside `Kramdown::Utils::StringScanner` (`skip_until` / `pos` were reported "undefined" on the SHADOW; three sites) and concurrent-ruby's `@backend` ivar-write mismatch (a `::Hash` read as `Concurrent::Hash`). The one addition, mail `lib/mail/yaml.rb:26` `wrong-arity` on `::YAML.safe_load(yaml, permitted_classes)`, is the honest report once `::YAML` reaches Psych instead of `Mail::YAML`: against the Ruby 4 Psych signature the positional form IS a two-argument call to a one-positional method. The line is the dead arm of a `Psych::VERSION < 3.1` guard, which is a general version-guard limitation, not this change's — filed as #627 (dead-arm pruning for literal version comparisons).
- **Critical review: APPROVE** (summarised in the description; the `is_a?` nit taken; residues filed as #625 / #637 / #638 / #639 / #626).

PR #646, opened 2026-09-02
Key rooted declarations by their de-rooted name across the sibling Rails plugins

### LINKED ISSUE #621: Rooted-key rendering + find/find('::…') retry survives on six sibling plugins' own indexes (#583's bug class)

zonuexe:
Fixes #621. #624 fixed the rooted-key bug for the shared `:model_index` fact; the same bug class survived on six sibling plugins' OWN indexes, each with the rooted `"::#{parts.join("::")}"` rendering plus a `find(name) || find("::#{name}")` retry that could not reach a de-rooted key. Each producer now renders declaration keys de-rooted (`class ::Foo` → `"Foo"`, and a rooted declaration nested in a module resets the lexical chain instead of producing `"Admin::::Foo"`), each index MERGES a re-opened class into one entry instead of letting the last file in the glob clobber it, and every retry arm is deleted.

Per plugin, with the mirror bugs the audit turned up:

- **rigor-actionmailer** — de-rooted keys, superclass and `mailer_base_classes` de-rooted, concern-module actions unioned; a reopen no longer blanks the earlier declaration's actions (a false `unknown-action` on the parent lib).
- **rigor-activejob** / **rigor-sidekiq** — same shape; the de-rooted query also repairs `#prepare`'s `known?` intersection, so a rooted job stops being dropped from `:reachability_roots`. Sidekiq additionally did not recognise `include ::Sidekiq::Job` as a worker at all. When two declarations both spell `#perform` with different shapes the envelope WIDENS rather than pinning either (glob order is not load order, ADR-5) — pinning is what produced a false `wrong-arity` on the parent lib.
- **rigor-actioncable** — actions and literal streams unioned, `dynamic_streams` OR'd; `< ::ApplicationCable::Channel` went undiscovered and drew `unknown-channel` (an ERROR) on correct code.
- **rigor-activestorage** — attachment rows unioned by name; the `dynamic_return` `receivers:` gate IS the key set, so a rooted owner never reached the rule at all.
- **rigor-actionpack** — the nested rooted case mis-keyed in both the discoverer and the analyzer's identical copy (both sides must render the same key); and the bigger fix, controller entries were ASSIGNED, so a controller declared in two files lost the other's `def`s and `before_action :that_method` fired a false `unknown-filter-method` (an ERROR).

Manifest versions bumped one minor step per changed plugin (a cached entry can be missing surface the merge restores). Specs: 30 new examples across the six integration spec files — per plugin a rooted-declaration regression guard, a reopen-keeps-the-earlier-surface discrimination case, and a must-still-fire sibling proving the merge widens the accepted set without silencing the rule. Discrimination: 3/18 actionmailer, 3/14 activejob, 4/24 sidekiq, 2/22 actioncable, 4/18 activestorage, 4/72 actionpack red on the parent lib; two of those reds are live false positives on correct code (the `wrong-arity` and `unknown-action` above).

Recorded, not fixed: actionpack loses a parent's rootedness through `qualified_name_for` (degrades to silence, FP-safe); `class Foo::Bar` inside `module Admin` keys its body's nested declarations inconsistently (pre-existing, would shift real keys); actionmailer's merged `missing_views` and actioncable's `action_methods` unions have no consumer today.

Draft pending the landing gates (corpus arms + CI).


zonuexe:
Landing gates:

- **Corpus FP diff** (full 11-target set, base = master arms at this branch's base, `--no-baseline --no-cache`): **+4 / −0, every addition an `info`-severity `worker-call` recognition trace on mastodon** — `DistributionWorker.perform_async` sites that the sidekiq plugin could not see before (`include ::Sidekiq::Job` went unrecognised as a worker), now matched against `#perform (arity 1..2)` with no diagnostic. No warning-or-above change on any target.
- **Verification:** the six plugin integration spec files green on the rebased tree (176 examples), rubocop clean on all 30 changed files, the scoped self-check clean over the six plugin `lib/` trees, and a CLI fixture run confirming the merge end-to-end. Discrimination per plugin recorded in the description; two of the parent-lib reds are live false positives on correct code (a `wrong-arity` from a pinned `#perform` and an `unknown-action` from a clobbered reopen).

PR #647, opened 2026-09-02
Treat the dead arm of a decidable version guard as unreachable

### LINKED ISSUE #627: Diagnostics inside the dead arm of a version guard (Gem::Version / RUBY_VERSION comparisons) are false positives on the target Ruby

zonuexe:
Fixes #627. A multi-version gem selects an arm with `RUBY_VERSION`, `Gem::Version`, `RUBY_ENGINE` or `RUBY_PLATFORM`, and the arm that cannot run on the Ruby being checked is dead code — a diagnostic there fires on a program that works. `mail/lib/mail/yaml.rb` is the corpus case surfaced by #614's arms: `::YAML.safe_load(yaml, permitted_classes)` is the Psych < 3.1 positional form, honest against the Ruby 4 signature and unreachable on Ruby 4.

`Inference::VersionGuard.verdict` is a pure function of the AST that folds only decidable comparisons — both sides literal or a readable version constant — and answers `:truthy` / `:falsey` / nil (undecidable, both arms stay live, the pre-existing behaviour). Three readers ask it, so they cannot disagree about which arm is dead: `StatementEvaluator` skips evaluating the dead arm (its writes never join the post-`if` scope), `ExpressionTyper#constant_predicate_polarity` folds the ternary form, and a post-pass drops the diagnostics whose source range lies inside a dead arm — the second half is needed because the rule walk visits every node whether or not the evaluator typed it, so a call with a self-typeable receiver still reported. `suppression.*` is deliberately applied AFTER the filter: a malformed `# rigor:disable` marker is an authoring error whether or not its code runs. `unless` arms are swapped, one-line ternaries drop only the dead half (ranges compared as `[line, column]` pairs), and the walk does not descend into a dead arm.

Binding docs updated in the same commit: `docs/type-specification/control-flow-analysis.md` (the fold's normative statement) and ADR-47 (the reachability decision, WD5).

Caveat for review: this branch's implementation agent was cut off by an API spend limit before it finished its own verification sweep, so the discrimination proof in its report is incomplete. The coordinator re-ran the spec files (469 examples green, rubocop clean) and the corpus arms; the arms and CI are the evidence carried here. Draft pending them.


zonuexe:
Landing gates:

- **Corpus FP diff** (full 11-target set, base = master arms at this branch's base, `--no-baseline --no-cache`): **+0 / −1**, and the one removal is exactly the diagnostic this change exists for — mail `lib/mail/yaml.rb:26` `wrong-arity` on `::YAML.safe_load(yaml, permitted_classes)`, the Psych < 3.1 positional arm behind `Gem::Version.new(Psych::VERSION) >= Gem::Version.new('3.1.0.pre1')`. Nothing else in the corpus moved: no other target's guards are decidable in a way that carried a diagnostic.
- **Verification:** 469 examples green across narrowing, remaining-rule-families, statement-evaluator and the changelog gate on the rebased tree; rubocop clean on the six changed files.
- **No independent adversarial review, and the implementation agent was itself cut off** by an API spend limit before finishing its own discrimination proof. The coordinator read the full diff (a pure `VersionGuard.verdict` shared by the evaluator, the ternary typer and the diagnostic filter; `unless` arms swapped; one-line ternaries split by `[line, column]`; `suppression.*` deliberately applied after), re-ran the specs and the corpus arms. That is a self-review plus CI, not the usual second instrument — flagged here so it is on the record, and the corpus effect above is the strongest evidence carried.

Issue #644, opened 2026-09-02
Cross-file value constants never resolve: FOO = :sym in one file reads Dynamic[top] from another


zonuexe:
Measured by the #622 implementation (branch `constant-absence-edge`, repro dirs under its scratchpad `p2` / `p5`), pre-existing on master:

- `FOO = :sym` declared in `b.rb`, read as `FOO` from `a.rb` → `Dynamic[top]`.
- `module Rails; LOG_LEVEL = :debug; end` in `b.rb`, read as `Rails::LOG_LEVEL` from `a.rb` → `Dynamic[top]`.

Cross-file VALUE constants never resolve: `in_source_constants` is consulted per file, not project-wide, so only same-file constant assignments type; class/module declarations (and the constant-assigned `Data.define` / `Struct.new` forms, which register in the discovery pre-pass's class sources) do resolve cross-file. This is why #622 could close the constant-absence staleness with the existing `class:Name` negative kind alone — there is no live staleness for value constants because they never resolved in the first place. It is also a plain precision gap on every project that keeps configuration in constants (`DEFAULT_LIMIT = 50` in one file, read in ten).

Fix direction: publish per-file constant ASSIGNMENTS (literal or foldable rvalue only — FP-first: an unfoldable rvalue publishes `Dynamic`, never a guess) into a project-wide table in the discovery pre-pass, keyed by qualified name, with the ADR-46 positive edge on a hit and a `constant:Name` negative kind (new — `appeared_symbols` must then emit it for a new assignment) on a miss; reassignment across files (two files assigning the same constant) → `Dynamic`. Gate: the two repros above type `:sym` / `:debug`; a reassigned constant stays gradual; the incremental add/remove sequence re-checks readers (snapshot SCHEMA bump if the row grammar grows).

PR #648, opened 2026-09-02
Record the absence edge an unresolved constant read leaves

### LINKED ISSUE #622: Incremental: an unresolved constant receiver records no absence edge, so declaring the constant later leaves its readers stale

zonuexe:
Fixes #622. An unresolved constant read left no ADR-46 dependency edge at all: `read_missing(:method, …)` fires only once the receiver constant has RESOLVED and the method lookup misses, so `Rails.logger` with no `module Rails` anywhere short-circuited to `Dynamic[top]` before either. Declaring `Rails` in a later-added file never re-checked the reader: a warm `--incremental` run kept `Dynamic[top]` while `--no-cache` answered the project reader and its true `call.undefined-method`.

`ExpressionTyper#unresolved_constant_fallback` — the single point where both `ConstantReadNode` and `ConstantPathNode` observe resolution answering nil — now records the edge (`record_missing_constant`), guarded by the recorder's `active?` integer read so a normal run pays nothing. The key reuses the EXISTING `class:<Name>` negative kind rather than a new `constant:` kind: `Incremental.appeared_classes` already reports a newly declared class/module by qualified name and `negative_affected` matches by simple name, so there is no new producer, no payload row, and no snapshot `SCHEMA` bump (a snapshot from an engine without the edge is already dropped by the fingerprint's engine-source part). One row per reference, keyed on the path's last segment (`::Rails` and `Rails` produce the same key; `Rails::Foo` → `class:Foo`), bounded by the accumulator's Set. Deliberate over-invalidation, pinned as a spec: a nested `module MyApp; module Rails` re-checks a reader of top-level `Rails` — the sound direction, the same grammar the override-ancestor negatives use. ADR-46's premise "a missed constant produces no diagnostic, so nothing goes stale" was wrong and is corrected in place; slice 3 records the third closure.

Specs (`Analysis::IncrementalSession`): the issue's repro (declared by an EDIT to an existing file), a qualified constant declared by a NEW file, a constant-assigned `Data.define` class (guards the "last segment always appears" premise), an unrelated class appearing → NOT re-checked, the nested over-invalidation; all five assert the merged diagnostics equal a full run. Discrimination: 4 of 5 red on the parent lib. CLI: the issue's sequence now byte-matches the `--no-cache` oracle, also under `--workers=2`. Recorded: `--verify-incremental` is structurally blind to this class (its subset re-analyses the consumer itself); cross-file VALUE constants never resolve at all (pre-existing, filed as #644), which is why no `constant:` kind is needed today.

Corpus arms: +0 / −0. Draft pending the review and CI.


zonuexe:
Landing gates:

- **Corpus FP diff** (full 11-target set, base = master arms at this branch base, `--no-baseline --no-cache`): **+0 / −0.** The change records an incremental dependency row; a `--no-cache` arm cannot differ, and it does not.
- **Verification:** the branch's spec files green on the rebased tree; rubocop clean; the discrimination proof recorded in the description.

zonuexe:
Landing gates:

- **Corpus FP diff** (full 11-target set, base = master arms at this branch base, `--no-baseline --no-cache`): **+0 / −0.** The change records an incremental dependency row; a `--no-cache` arm cannot differ, and it does not.
- **Verification:** the branch's spec files green on the rebased tree; rubocop clean; the discrimination proof recorded in the description.

Issue #645, opened 2026-09-02
Straight-line content mutation on a Union receiver records nothing: out = flag ? 5 : [2]; out << 2 keeps the Tuple's arity


zonuexe:
Found by the #631 implementation (branch `union-dynamic-arm-survives-block-seam`), pre-existing on master and adjacent to #631 (which fixed the block/loop seam, not the straight-line side):

`MutationWidening.widen_for_mutator` declines a `Union` receiver outright, so a straight-line content mutator on a union seed records NO mutation evidence — no arity-forget, no added element:

```ruby
def f(flag)
  out = flag ? 5 : [2]
  out << 2
  Rigor.dump_type(out)   # 5 | [2]   — the runtime value on the array branch is [2, 2]
end
```

The `[2]` Tuple keeps its arity after a mutation falsified it — an unsound retained pin on the straight-line side (a later `out.size == 1` or `out.last == 2` would fold on a value the program does not hold). The block twin (`[1].each { out << 2 }`) now answers `5 | Array[2]` after #631.

Fix: `widen_for_mutator` should map a Union memberwise — widen every member that is a carrier of the mutator's class (`Tuple` → `Nominal[Array, [elem | added]]`, `HashShape` likewise) and keep the others whole (`5 | Array[2]`), mirroring #631's `array_residue` / `hash_residue` partition so the two seams agree. Must-not-fire: the shape above with `out.size == 1` (no fold); must-still-succeed: a non-union Tuple receiver widens exactly as today; a Union of two Tuples widens both.

PR #649, opened 2026-09-02
Keep a Union seed's non-collection arms through the content seam

### LINKED ISSUE #631: Block seam drops a whole-variable Dynamic arm of a Union seed: out = flag ? u : [2]; [1].each { out << 2 } types Array[2] and fires on correct code
Found by the #615 implementation (branch `array-new-dynamic-size-seed`), reproduced byte-identically on master `926ce24b` — pre-existing, and a FIRING false positive rather than the imprecision #615's text suggested. Distinct from #586/#616 (a Dynamic ELEMENT arm inside an Array carrier): this is a whole-variable Dynamic arm of a Union.

```ruby
def f(flag, u)
  out = flag ? u : [2]     # Dynamic[top] | [2]
  [1].each { out << 2 }    # -> Array[2]; the Dynamic arm is gone
  out.first.upcase         # error: undefined method `upcase' for 2   (FP on correct code when u is a String array)
end

def g(flag, u)
  out = flag ? u : [2]
  [1].each { out << 2 }
  puts "hit" if out.first == 3   # warning: condition is always falsey  (FP)
end
```

The straight-line sibling (`out << 2` with no block) keeps `Dynamic[top] | [2]` and is quiet, so this is specific to the block seam. Root: `ContentJoin.join_array_content` builds a fresh `Nominal[Array, [union]]` and REPLACES the whole binding; a Union pre-state member that is a whole-variable `Dynamic` (not an Array-ish carrier) contributes no element evidence through `collection_element_types`'s `else []` branch and then vanishes.

Fix direction (FP-first): the block seam's rederived carrier must keep every non-collection member of the pre-state Union alongside the rederived Array — `Dynamic[top] | Array[2]` here — the same "a seed's own gradual arm survives" rule ADR-56 WD2.9 states for the element arm, lifted to the whole-variable arm. Must-not-fire: both shapes above; must-still-succeed: a pre-state Union of two Array carriers still joins to one Array; a `nil | [2]` pre-state (the `||=`-style seed) keeps its nil arm only if the block can run zero times (it can — `each` over a possibly-empty receiver), which is the existing behaviour to pin. Also from the same review: a "`while` inside a block appending a body-local reads `Array[Dynamic[top]]`" claim from #615's text did NOT reproduce in five shapes (every one reads identically to its top-level `while` twin); the only `Array[Dynamic[top]]` came from appending a genuinely untyped block parameter, which is honest — drop that item from #615 unless a concrete repro appears.

zonuexe:
Fixes #631. The block and loop content seams rederive ONE collection carrier and write it over the whole binding. That is right only for the pre-state members the mutation applied to as a collection of that class: `ContentJoin.collection_element_types` / `hash_shape_key_values` answer nothing for a whole-variable `Dynamic`, a foreign `Nominal` or a `Constant`, so those contributed no element evidence and were then overwritten out of existence — `out = flag ? u : [2]; [1].each { out << 2 }` typed `Array[2]`, and `out.first.upcase` fired `undefined method 'upcase' for 2` / `out.first == 3` folded always-falsey on correct code (the straight-line twin kept `Dynamic[top] | [2]`).

`join_array_content` / `join_hash_content` now PARTITION the pre-state: a member readable as a carrier of the target class (`Tuple`, `Nominal[Array]`/`Nominal[Hash]` with or without type args, a `Difference` through its base, a `Union` memberwise) is absorbed into the rederived carrier; every other member — whole-variable `Dynamic`, foreign `Nominal`, `Constant`, `Refined`, a `Difference` whose base is not a carrier — survives whole beside it, each examined exactly once (`array_residue` / `hash_residue`, written recursion-for-recursion against the element readers so the absorbed set and the residue cannot drift). The loop seam shares the join and gets the rule for free (its twin fired the identical FP); `MutationWidening`'s straight-line caller only ever hands the join a `Nominal`, so it is byte-identical. Decision, measured before taking it: `nil` does NOT survive — the mutation refutes it (`NilClass` has no content mutator), the zero-iteration path is modelled upstream (the `while` base scope's nil injection, slice A's `Constant[nil]` seed), and keeping it added a `possible-nil-receiver` to the deliberate `r = nil; while …; r ||= []; r << x; end; r.each` idiom that master is quiet on — both `nil`-seed shapes are byte-identical to master. Two Array carriers still join to one Array; #615's bare `Nominal[Array]` is a carrier and still closes; #616's element-arm rule operates on the absorbed side only (no double-widening). Recorded in ADR-56 WD2.11 and normatively in `docs/internal-spec/inference-engine.md`. A second, separately droppable commit removes a stray `|||||||` diff3 marker that reached master inside ADR-56 (from #620's conflict resolution).

Specs: content_join_spec (Dynamic member kept beside the Array; foreign Nominal kept; two carriers collapse; bare `Nominal[Array]` absorbed; nil dropped, pinning the decision; the Hash twin) + a new integration fixture `union_seed_residue.rb` (both seams, the always-falsey shape, the must-fire nil control). Discrimination: 6 examples red on the parent lib (the fixture's two FP lines and the always-falsey line included). Residue filed: #645 (the straight-line `widen_for_mutator` declines a Union outright — the Tuple keeps a falsified arity).

Corpus arms: +0 / −0. Draft pending the review and CI.


zonuexe:
Landing gates:

- **Corpus FP diff** (full 11-target set, base = master arms at this branch base, `--no-baseline --no-cache`): **+0 / −0.** No corpus site mutates a Union-seeded local (a whole-variable `Dynamic` or a foreign class beside a collection literal) inside a block or loop and then reads it into a firing diagnostic on either arm; the fixed shape is the wrong-precise family the #615 review found.
- **Verification:** the branch's spec files green on the rebased tree; rubocop clean; the discrimination proof recorded in the description.

Issue #643, opened 2026-09-02
Nested literal containers keep their pins through an indexed mutation: [flag ? [1] : [2]][0] << 5 then .last == 5 folds always-falsey


zonuexe:
Residue left by the #615 review (byte-identical on master; the constructor-side twin was fixed on the #615 branch by widening `Array.new`'s fill/block element memberwise). The plain literal keeps the hazard:

```ruby
c = [flag ? [1] : [2]]   # Tuple[[1] | [2]] — both members are pinned tuple literals
c[0] << 5
puts "hit" if c[0].last == 5   # warning: flow.always-truthy-condition: always falsey  (FP — it prints)
```

A Union of literal containers as an element keeps each member pinned, so a later mutation through an rvalue temp (`c[0] << 5`) cannot retract the pin and the comparison folds. The non-union literal `[[1]]` reads the same way (`[[[]]]` → `b[0][0] << 1; b[0][0].first` → `nil`, also on master) — the rvalue-temp mutation hazard is engine-wide for nested literal containers (#601's aliasing umbrella is the design home: mutation evidence updates exactly one binding, and here the mutated binding is the temporary `c[0]`).

FP-first direction: an element read through an index (`c[0]`, `c.first`) that is then MUTATED should widen the element's pin one step in the container (`Tuple[[1] | [2]]` → `Tuple[Array[Integer]]`), or the comparison fold should decline when the operand's pin comes from a nested literal reachable through a mutable read. Must-not-fire: the snippet above and the `[[[]]]` twin; must-still-succeed: `c = [[1]]; c[0].last == 1` (unmutated) still folds truthy.

PR #650, opened 2026-09-02
Seed a real element arm for a dynamic-size Array.new

### LINKED ISSUE #615: Array.new(n) types as elementless Array, so the block/loop seams close it over stores: wrong-precise after the #586 rule

zonuexe:
Fixes #615. `Array.new` with a non-literal size answered a bare `Array` — a nominal with no element arm — which the ADR-56 block/loop seams read as an ELEMENTLESS seed and closed over the block's stores: `acc = Array.new(n); [1].each { acc.push(1) }; acc.first == 1` folded always-truthy (the runtime array holds `n` untouched slots), and `acc = Array.new(n, "x"); [1].each { acc.push(1) }; acc.first.upcase` reported `undefined method 'upcase' for 1` on correct code.

The fix is in the constructor, not the seam: `array_new_lift` falls through to `array_new_seed`, which returns a real `Nominal[Array, [E]]` so the seams see a declared-like carrier and keep its arm under #616's rule. Element choices, all FP-driven: the fill value's class (value-pin widened, as `hash_new_lift` does — slots are rewritten over the array's lifetime), a literal container fill / block result widened one step (`Array.new(n) { [] }` builds `n` independent arrays the program appends to — `Array[Tuple[]]` would read `adj[i].first` as `nil`), a Union fill/result widened memberwise (review round 2), and — the review's first-round blocker — the no-fill form stays GRADUAL rather than `Array[nil]`: a Nominal seed gets declared-carrier semantics (the join keeps the seed arm forever and the straight-line mutators never widen a Nominal's element), so a placeholder `nil` becomes a permanent nil possibility over the allocate-then-fill idiom (`dp = Array.new(n); dp[0] = 1; (1...n).each { |i| dp[i] = dp[i-1] + xs[i] }` drew `undefined method '+' for nil`). Literal-size `Array.new(3, 0)` keeps its tuple fold; zero-arg `Array.new` stays elementless (it really is empty). ADR-56 WD2.9's parenthetical corrected and a new WD section records the rule, the element choices, why `nil` is not seeded, and the recorded-not-fixed fill-class-survives-store note; no binding statement of the constructor fold exists in type-specification / internal-spec (verified by the review).

**Review:** REQUEST_CHANGES (the two blockers above, fixed in the third commit — the reviewer's probe files now show only pre-existing reds: the unmutated literal-size fold, the declared `Array[nil]` control, and the plain-literal twin filed as #643), then delta re-check. Integer-gate membership, the copy overload (`Array.new(other_array)` → gradual, never `Array[nil]`), fill/block forms (`i * 2`, `i.to_s`, `SomeClass.new`, unions, `raise`, `&:to_s`), the seams (block, `while`, nested, index-write, the sieve `Array.new(n, true)` flipped in a block — parent folded always-truthy) all verified. Discrimination: 12 of 401 examples red on the parent lib (the integration fixture's FP lines 21/33/46 plus the DP/`replace` lines 69/90 on the pre-review lib); self-check over every `lib/` file using `Array.new` byte-identical on both arms.

**Corpus arms (full 11):** +0 / −0. Draft pending CI.


zonuexe:
Landing gates:

- **Corpus FP diff** (full 11-target set, base = master arms at this branch's base, `--no-baseline --no-cache`): **+0 / −0** — after one fix the arms themselves forced. The first collection showed a NEW error on concurrent-ruby, `undefined method 'result' for nil` at `promises.rb:2025`, from `@Resolutions = ::Array.new(count, nil)` filled by index and then read: the explicit-placeholder twin of the no-fill blocker the review had already adjudicated, since a nil-only element is equally permanent on a `Nominal` carrier. A nil-ONLY fill now seeds the gradual arm as well (a fill that is only partly nil, `flag ? nil : "s"`, keeps both arms — the guard reads nil-only, not nil-bearing); the rule and its rationale are in ADR-56's new section, with a unit example either side and a fixture case mirroring the concurrent-ruby shape. Re-collected arms: clean.
- **Reviews:** one full adversarial round (REQUEST_CHANGES: the no-fill `Array[nil]` seed and the un-widened union fill), then a delta round (REQUEST_CHANGES: nested containers one level deeper), both fixed and re-verified; the reviewer's probe files end with only pre-existing reds. The third finding above came from the corpus gate rather than a reviewer — the two instruments caught different halves of the same family, which is the case for keeping both.
- 520 examples green across the constructor, seam and widening spec files on the rebased tree; rubocop clean; the self-check over every `lib/` file using `Array.new` byte-identical to master.

PR #779, opened 2026-09-05
Drop comments that restated a nearby contract

zonuexe:
Stricter keep test: a comment stays only if it states a constraint the next lines do not. ADR citations, fail-soft `nil`, false-positive bounds, and declined alternatives stay.

Dropped:

- Copy-paste that restated a nearby contract.
- Bare YARD `@param` / `@return` that only repeated the signature.
- Heuristic banners that restated the method name (`h1_activesupport` does not need `--- H1 — likely ActiveSupport ---`).
- Catalog labels that restated the symbols they sit on (`:hash, :==, :eql?` is already identity hashing).
- `Returns the X` openers on Reflection facades; the nil / collision / dispatcher sentences remain.

Remaining type-bearing comments are rbs-inline (`# @rbs name: T -- prose`). This tree already declares those methods in `sig/`, so `.rigor.dist.yml` keeps the magic-comment gate: lib/ comments are the dialect, `sig/` is the type source. Unannotated siblings no longer get `untyped` skeletons, and a method already in `sig/` keeps the `.rbs` declaration.

The comment rewrite is internal hygiene. The ingestion guards are in `changelog.d/fixed/comment-redundancy.md`.

### DIFF
--- a/.rigor.dist.yml
+++ b/.rigor.dist.yml
@@ -2,6 +2,12 @@
 target_ruby: "4.0"
 paths:
   - lib
-plugins: []
+# Types for lib/ live in sig/. Listing the plugin here skips ADR-93 auto-wire
+# (magic-comment-free) so `# @rbs` comments in lib/ are not a second type source.
+plugins:
+  - gem: rigor-rbs-inline
+    id: rbs-inline
+    config:
+      require_magic_comment: true
 cache:
   path: .rigor/cache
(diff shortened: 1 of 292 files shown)


zonuexe:
Superseded by #822: the position on types in comments was redefined — a type Rigor does not check is not written at all, so the rbs-inline rewrite and the ingestion gate are withdrawn. The local rebase of this branch is kept as `backup/comment-redundancy-pre-rebase`; the two engine behaviours this PR introduced (dropping unannotated skeletons; `sig/` silently winning over an inline duplicate) are filed as separate issues.

Issue #785, opened 2026-09-07
HKT directive / sugar-scan errors are a silent no-op against the production reporter


zonuexe:
> *This was generated by AI during triage.*

Found during the #776 / #783 review.

`Rigor::RbsExtended::HktDirectives#record_hkt_error` (`lib/rigor/rbs_extended/hkt_directives.rb:289-296`) routes every HKT directive failure — malformed `hkt_register` / `hkt_define`, arity mismatch, bad variance, unparseable body — through:

```ruby
if reporter.respond_to?(:record)
  reporter.record(directive: "hkt", message: message, source_location: source_location)
elsif reporter.respond_to?(:<<)
  reporter << { directive: "hkt", message: message, source_location: source_location }
end
```

The production reporter is `Rigor::RbsExtended::Reporter` (`lib/rigor/analysis/runner.rb:334`), whose surface is `record_unresolved` / `record_lossy_projection` — **neither `record` nor `<<`**. So every `record_hkt_error` call is silently dropped in a real run. The `scan_rbs_loader` doc comment ("the reporter records an `:info` entry") does not hold for HKT.

The new per-alias fail-soft skip added in #783 inherits the same silence — a malformed vendored `type` alias is skipped with no diagnostic.

Fix direction: give HKT directive/scan failures a real `dynamic.rbs-extended.*`-style `:info` diagnostic. `ready-for-human` because it needs a diagnostic identifier + a `docs/type-specification/` entry (the spec binds), not just a wiring change.

Issue #776, opened 2026-09-05
0.3.7: HktSugarTranslator passes name_scope: to RbsTypeTranslator.translate, which does not accept it — every file fails with "internal analyzer error"


n-rodriguez:
## Context

- rigortype **0.3.7** (0.3.6 is fine — see below), Ruby 4.0.6, rbs 4.2.0
- Rails 8.1.3 application, `rbs collection install` from `ruby/gem_rbs_collection` (91 gems)
- `plugins:` and `signature_paths:` are irrelevant here (both eliminated by bisection)

## Observed

Every single file analysed fails, with no analysis performed at all:

```
$ rigor check app/models/camping.rb
app/models/camping.rb:1:1: error: internal analyzer error: ArgumentError: unknown keyword: :name_scope

1 error(s) in 1 file(s)
```

Same on `app/datatables/application_datatable.rb`, `lib/concerto/events.rb` — the file does not matter.
On 0.3.6, the same command on the same project reports `No diagnostics`.

## Cause

`HktSugarTranslator#fallback_to_type_leaf` passes a keyword the callee does not declare.

`lib/rigor/inference/hkt_sugar_translator.rb:79-86`:

```ruby
def fallback_to_type_leaf(type)
  require_relative "rbs_type_translator" unless defined?(RbsTypeTranslator)
  rigor_type = RbsTypeTranslator.translate(
    type,
    alias_expander: nil,
    type_vars: {},
    name_scope: @name_scope,     # <-- not a parameter of the callee
    self_type: Rigor::Type::Combinator.untyped,
    instance_type: Rigor::Type::Combinator.untyped
  )
  HktBody::TypeLeaf.new(type: rigor_type)
end
```

`lib/rigor/inference/rbs_type_translator.rb:114-115`:

```ruby
def translate(rbs_type, self_type: nil, instance_type: nil, type_vars: EMPTY_TYPE_VARS,
              alias_expander: nil)
```

There is no `name_scope:` parameter, so any call reaching this fallback raises.

## Backtrace

Captured with a `TracePoint(:raise)` probe, since the CLI rescues it:

```
lib/rigor/inference/rbs_type_translator.rb:114:in 'Rigor::Inference::RbsTypeTranslator.translate'
lib/rigor/inference/hkt_sugar_translator.rb:81:in 'Rigor::Inference::HktSugarTranslator#fallback_to_type_leaf'
[...]

Issue #665, opened 2026-09-02
Absence-only specs pass when a check rule raises: an internal analyzer error erases every other diagnostic and the harness does not notice


zonuexe:
Found by the adversarial review of [#632](https://github.com/rigortype/rigor/issues/632), where it caused a vacuity check to pass for the wrong reason during the review round itself.

## What happens when a check rule raises

`lib/rigor/analysis/runner.rb:1736-1746` (and `worker_session.rb:229-231`) rescue `StandardError` around `analyze_file` and return **one** diagnostic for the whole file:

```
Diagnostic.new(path:, line: 1, column: 1, message: "internal analyzer error: <class>: <msg>", severity: :error)
```

with `rule: nil`. Every other diagnostic that file would have produced — real findings and `dump.type` infos alike — is discarded. Other files continue.

In the **CLI** this is loud and correct: `Result#success?` is `diagnostics.none?(&:error?)`, so the run exits 1 and `make check` / CI go red.

```
code.rb:1:1: error: internal analyzer error: NameError: reviewer-injected boom
1 error(s) in 1 file(s)
exit=1
```

## The hazard is in the spec harness

`spec/integration/support/plugin_helpers.rb`'s `run_plugin` / `run_plugin_in_dir` are used by specs that assert **absence** — "does not fire `call.undefined-method`", "witnesses no diagnostic". Those assertions pass when the rule crashed, because:

- `rules(result)` maps `qualified_rule`, which is `nil` for an internal analyzer error, so `expect(rules(result)).not_to include("call.undefined-method")` holds;
- `expect(dumps(result)).to all(eq(...))` passes vacuously on the empty list an internal error leaves behind.

Reproduced by injecting `raise NameError, "boom"` into a check-rule helper: the plugin spec reports `2 examples, 0 failures` while the analyzer was in fact producing nothing but an internal error.

This is not hypothetical. During #632's implementation the first version of a gate raised `NameError` at runtime, and the spec written to prove the gate worked passed anyway — the defect was found only because the reviewer re-verified on a mutated copy rather than trusting the tally.

## Fix

**Fail fast in the harness.** No spec should legitimately produce an `internal analyzer error` diagnostic, so `run_plugin` / `run_plugin_in_dir` (and any sibling helper that returns a `Result`) should raise when one appears, naming the file and the underlying exception. That converts the whole class of "absence assertion passes because the analyzer crashed" into a loud failure at the exact spec that hid it.

[...]

Issue #696, opened 2026-09-03
A DuplicatedMethodDefinitionError leaves the class known-but-empty and the run exits 0: make the build failure reportable


zonuexe:
Raised and scoped by the [#672](https://github.com/rigortype/rigor/issues/672) work, which fixed one route into this failure and deliberately did not fix the failure itself.

An `RBS::DuplicatedMethodDefinitionError` during a definition build leaves the class **known but empty**. A warning is printed; the run still **exits 0**, and every call on that class silently answers `Dynamic[top]` — including calls that should report.

## The repro that makes the case

A **vendored copy** of a bundled plugin's `.rbs` in a project's `sig/`, alongside the gem overlay (#672's path-based stand-down deliberately does not catch a copy):

```
$ rigor check --workers=4
rigor: RBS definition build failed for `…`      × 2709
code.rb:1:7: info: dump_type: Dynamic[top]      # Time.current
                                                # Time.no_such_method_at_all — nothing reported
##### exit=0
```

The copy reopens `Object`, so **every class that inherits it collapses — the whole bundled type universe** — and the run is green. `--workers=0` prints the failure once; the 2709 is the per-loader warn memo re-firing per fork worker (adjacent to the known [#295](https://github.com/rigortype/rigor/issues/295)).

## Why this is the family that keeps recurring

This is the same shape as [#665](https://github.com/rigortype/rigor/issues/665) / [#674](https://github.com/rigortype/rigor/issues/674) / [#683](https://github.com/rigortype/rigor/issues/683) / [#686](https://github.com/rigortype/rigor/issues/686): **an internal failure that makes the analyzer report LESS while still exiting 0.** Every "nothing fires" assertion passes on it, in the suite and in a user's CI alike. #672's own spec found one of its examples passing vacuously on master for exactly this reason.

## The proposed answer, with its rationale

A third `rbs.coverage.*` sibling — `rbs.coverage.definition-build-failed` — authored `:warning` and promoted to `:error` by the existing `reject-unparseable-signatures` bleeding-edge feature (`lib/rigor/bleeding_edge.rb:116-117`). That is the ladder `rbs.coverage.quarantined-signature` and `rbs.coverage.environment-build-failed` already sit on (`docs/type-specification/diagnostic-policy.md`), and it slots between them by consequence: a quarantined file removes what one file declared, this removes what one **class** declared, an env-build failure removes everything.

[...]

PR #787, opened 2026-09-07
Collapse repeated internal-error rows into one run-level report

### LINKED ISSUE #784: analyze_body reports a shared-infra crash as a per-file "internal analyzer error"
> *This was generated by AI during triage.*

Split out of #776 (fixed in #783).

`Rigor::Analysis::WorkerSession#analyze_body` (`lib/rigor/analysis/worker_session.rb:230-231`) wraps the whole per-file pipeline in `rescue StandardError => e` and emits `analyzer_error(path, "internal analyzer error: ...")`.

When the exception originates in **shared, memoised infrastructure** rather than in per-file analysis — the HKT registry build (#776), the RBS env build, a plugin overlay — every file in the run gets an identical bogus diagnostic and the exit is non-zero, but **nothing states that analysis never actually ran**. A project goes from hundreds of real diagnostics to exactly one-per-file, which reads like "almost clean" rather than "totally broken". #776 was the worst case: a single malformed `type` alias in one collection gem.

This is a judgment call, hence `ready-for-human`:

- A distinct run-level surface ("analysis aborted before it started: <cause>") for failures in shared setup, vs. genuine per-file analyzer errors.
- Or: detect that the same `internal analyzer error` fired on *every* target and collapse it to one run-level error.
- Interaction with the pool: the shared build may run in the parent (`--workers=0`) or lazily in each worker (`--workers=N`); the surface has to say the same thing both ways (cf. #696).

Not blocking anything — #783 stops the specific crash — but the misleading-symptom problem outlives it.

zonuexe:
## What

Split out of #776. When a shared, memoised sub-build raises during per-file analysis — the HKT registry (#776), a plugin overlay, the RBS env, a Ractor `IsolationError` ([pool_coordinator.rb:299](https://github.com/rigortype/rigor/blob/master/lib/rigor/analysis/runner/pool_coordinator.rb#L299)) — every file's `analyze_body` `rescue StandardError` catches it and folds it into the **same** `internal analyzer error` row. The run comes back looking almost clean: N byte-identical rows, a non-zero exit, and nothing stating that the analyzer never ran. #776 produced one such row per file across a whole Rails app.

## Fix

Detect-and-collapse at the aggregator, the generic option — it catches every shared-build failure, present and future, without threading a new snapshot through each build seam.

**`Runner::DiagnosticAggregator#collapse_repeated_internal_errors`** — any group of ≥2 `:check_rule` rows with a byte-identical message folds to its first occurrence (kept in place as a sample) plus one `analyzer.internal-error` summary at `.rigor.yml:1:1` naming the count, the denominator, and the sample path. Wired on the miss path before the stream reaches the run cache, so a warm re-serve gets the collapsed form. A healthy run has no such group (no-op); a genuine per-file check-rule bug produces per-file-varying messages and is not collapsed. Authored **`:error`** — unlike its `:warning` `rbs.coverage.*` neighbours, a shared build raising is always a Rigor defect, never the user's `sig/`, so an `:error` default cannot turn a clean project red without a real bug.

**`HktRegistryHolder`** now memoizes a `StandardError` from its build and re-raises the original, instead of re-running — and re-raising — the failing build once per file.

**`CrashSignature.check_rule_message`** centralises the rescue message the two `analyze_body` twins build (#665 — they must not drift) and appends the first `lib/rigor/` backtrace frame, so the retained sample row points at the raise site.

The summary keeps the `internal analyzer error` prefix, so `CrashSignature` still classifies it `:check_rule` and `Result#crashed?` / the ADR-69 kill oracles / the spec-harness guard keep recognising the run as one whose analysis was discarded.

## Tests

- `spec/rigor/analysis/runner/diagnostic_aggregator_spec.rb` — `collapse_repeated_internal_errors`: no-op on a healthy run, no-op on a lone crash, no-op on per-file-varying messages, folds a multi-file group (sample kept by identity + summary appended, count / denominator / severity / rule / classification).
- `spec/integration/shared_build_crash_collapse_spec.rb` — a real 3-file run with a stubbed shared-build raise comes back with 1 sample + 1 summary, not 3 rows. Registered in the `spec_analyzer_guard_spec.rb` allowlist (its subject is the crash envelope, so `guarded_run` would raise before the collapse could be read).
- `spec/rigor/environment/hkt_registry_holder_spec.rb` — failure memoization, and that a non-`StandardError` does not poison the slot.
- `spec/rigor/analysis/crash_signature_spec.rb` — `check_rule_message` / `crash_frame`.

`make verify` + `make docs-check` green; `git diff --check` clean.

## Docs

- `docs/type-specification/diagnostic-policy.md` — new `analyzer.*` family row.
- `docs/internal-spec/inference-engine.md` (in #783) already carries the HKT scan contract; this adds `analyzer` to `NON_CHECK_DIAGNOSTIC_FAMILIES`.

## Not in this PR

#785 — `HktDirectives#record_hkt_error` is a silent no-op against the production reporter — is unrelated and stays open.

Closes #784


zonuexe:
Closing without merge — superseded.

An adversarial review found three high-severity consequences of the generic post-hoc collapse that are properties of the design rather than bugs to patch: (D1) message identity cannot distinguish a shared-build crash from a per-file rule bug, so a partial per-file crash was folded and the affected files vanished from every path-indexed consumer; (D2) the summary carries run-shape state (denominator, sample path), so `--verify-incremental` reported a phantom divergence between a subset run and the full run; (D3) `IncrementalSession#per_file` drops `.rigor.yml` rows on the premise that run-level rows are recomputed fresh each run — true for a seam-recorded row, false for a collapse of the per-file rows it destroys — so `--incremental` and the LSP project surface cached the crashed files as clean.

The replacement follows the pattern the repo already uses for every other shared-build failure (`env_build_failure`, `definition_build_failures`, #696): guard the seam at `Environment#hkt_registry`, record the failure, degrade to the pre-scan registry so analysis proceeds, and surface one `rbs.coverage.hkt-scan-failed` row from the aggregator. Per-file rows are untouched, so D1–D3 cannot occur. Full rationale on #784.

`CrashSignature.check_rule_message` / `crash_frame` from this branch carry over to the replacement.

Issue #775, opened 2026-09-05
Bring `rigor check lib` allocations back toward the v0.3.6 18.8M


zonuexe:
Blessed at the v0.3.7 cut so the release can land. The next cycle should recover the cost, not live with it.

## What moved

`make bench-perf` (`tool/bench.rb --target lib`) on Linux CI, in-process `rigor check --no-cache` of `lib`:

| cut | allocations | peak_rss_kb | wall_s | diagnostics | lib files / lines |
| --- | ---: | ---: | ---: | ---: | ---: |
| v0.3.6 (run 33312083066) | 18,849,919 | 315,508 | 11.612 | 1 | 427 / 92,261 |
| v0.3.7 (run 33944842772) | 36,171,454 | 417,716 | 25.266 | 2 | 444 / 100,804 |

Allocations +91.9% against a +5% band. RSS +32.4% against +10%. `lib` grew about 9% lines, so this is analysis cost, not target growth. Mastodon OSS sweep on the same release-gate run stayed inside its thresholds, so the extra work did not show up as a diagnostic flood on that corpus.

The extra diagnostic is a `def.return-type-mismatch` warning on `lib/rigor/inference/expression_typer.rb` after `Rigor::Type::Maybe` / `Rigor::Type::Result` joined the declared union.

The v0.3.7 `bench/baseline.json` refresh copies the CI artifact as-is (single run; no second sample to take the lower of wall/RSS). Follow the note in that file.

## Gate

A fix is done when a Linux `release-gate.yml` `make bench-perf` run reports allocations and RSS inside the bands of a baseline that is *not* this 36.2M blessing — either by recovering toward ~19M, or by a measured, explained remainder with a new baseline whose `note` says why the leftover is real.

Do not recalibrate again without a measurement of *which* v0.3.6..v0.3.7 change paid the extra 17M allocations. Compact-namespace nesting, callee rewalk, class-method ancestor walks, and the new bundled plugin RBS (FFI family, dry-monads) are the first suspects; none of those is confirmed.

## Out of scope

- Recalibrating for `wall_s` alone (ADR-50 WD4; `bench/thresholds.yml`).
- Widening `bench/thresholds.yml` to hide the 36.2M number.

zonuexe:
[...]

PR #819, opened 2026-09-08
Recover the v0.3.7 allocation regression: attribution plus fifteen mechanical levers

zonuexe:
Part of #775. Attributes the v0.3.6 → v0.3.7 allocation regression (`make bench-perf`: 18.8M → 36.2M on Linux CI) and lands the mechanical half of the recovery: `rigor check --no-cache lib` goes from 36,492,665 to 21,933,996 allocations (−39.9%) with diagnostics byte-identical to master on the `lib` self-check after every commit and on redmine (app + lib, seven plugins) at the end. The full measurement record is [`docs/notes/20260908-v037-allocation-regression-attribution.md`](https://github.com/rigortype/rigor/blob/perf-775-allocation-levers/docs/notes/20260908-v037-allocation-regression-attribution.md); the instruments are on the unmerged branch `perfbench-harness-775`.

## Where the 17M went

A per-merge sweep of the 127 code-touching first-parent commits, plus an A/B of the v0.3.6 engine and the current engine analysing the *same* tree, put the whole regression in per-file typing (`ExpressionTyper#type_of` +11.1M, `StatementEvaluator#evaluate` +3.7M); the one-time RBS environment build that dominated the July note is unchanged at ~2M, the 9% larger target is worth +1.7M, and the `rbs` 4.2.0 bump +0.2M. No single merge is the regression: #547 (per-parameter binder) +4.1M, #556 (alias translation) +2.3M, #753 (autoload-safe `resolve_class`) +1.8M, #664 (declared-reader precedence) +1.7M, #543 +0.9M, #712 +0.7M, then a long tail. The per-call story is `Rigor::Type::t`: a 21-member alias whose *translation* (not its expansion, which #556 memoised) was rebuilt at every call site and then flowed into every join — 13,592 21-member unions and 4.8M `Nominal#==` calls per run — plus class-name string churn, `[nil, nil]` / `{}` / `[]` answers, and lambdas, `zip`s and `dup`s on the overload-selection path.

## Levers (one commit each, cumulative)

| lever | allocations after | Δ |
| --- | ---: | ---: |
| master | 36,492,665 | |
| `normalize_name` copies only when a `::` prefix exists | 35,545,804 | −946,861 |
| `content_mutation_target` declines with one frozen pair | 34,598,075 | −947,729 |
| `Nominal` shares one frozen empty `type_args` | 34,187,174 | −410,901 |
| `RbsExtended` readers share their empty answers | 33,561,727 | −625,447 |
| translator memoises `TypeName#relative!.to_s` | 31,436,905 | −2,124,822 |
| translator memoises a closed alias expansion's translation | 30,856,415 | −580,490 |
| `union(x, x)` answers `x`; identity before `==` in `unique_members` | 30,278,276 | −578,139 |
| `resolve_class` memoises the split of the name | 28,484,075 | −1,794,201 |
| overload selection without lambdas, `**` splats, `zip`, `dup`, `compact` | 26,655,925 | −1,828,150 |
| receiver-affinity reorder only when the order changes | 25,304,174 | −1,351,751 |
| one `AcceptsResult` per literal reason | 25,055,693 | −248,481 |
| one `FactStore::Target` per local name | 24,322,534 | −733,159 |
| `class_type_param_names` memoised and frozen on the loader | 22,501,792 | −1,820,742 |
| empty type-vars map and discovered-lookup pair shared | 22,115,323 | −386,469 |
| `nil` / `true` / `false` Constants interned | 21,938,254 | −177,069 |
| head (lint follow-ups) | 21,933,996 | |

Every step is a shared frozen empty answer, a memo of a pure function, or a rewrite that computes the same thing without the intermediate object; none changes what is inferred. The one spec change pins `class_type_param_names`'s stronger contract (the memoised list is frozen, so a mutation attempt raises instead of corrupting the table).

## The remainder

Against v0.3.6's 18.8M the branch sits at about +16% rather than +92%. On the same target the v0.3.6 engine allocates 21.4M and this branch 23.6M (both under the phase probe), while this branch does 64% more user-method return inferences and 68% more body evaluations — the inference volume #547 / #556 / #664 buy. What is left is that volume plus design seams (a rebuilt `Scope` per binding, a `RuleWalk::Context` per node, an `ExpressionTyper` per `Scope#type_of`), not mechanical removals; the note lists the measured-and-left items.

## Verification

- `make verify` green (10,333 examples), `git diff --check` clean, `make docs-check` green.
- `rigor check --no-cache --no-stats --format json lib`: byte-identical to master after every commit.
- redmine (survey checkout, `app` + `lib`, seven plugins, `--no-cache`): 354,912 bytes of JSON, byte-identical between the master engine and this branch.
- `RIGOR_BUDGET_TRACE` memo counters identical to master back-to-back (infer entries 42,728, body evals 18,125); the union-arity histogram is the intended difference (unions ≥ 10 members 20,250 → 7,856).

## Follow-up (not in this PR)

The issue's gate is a Linux `release-gate.yml` `make bench-perf` inside the bands of a non-blessed baseline. The next release-gate run on `master` after this merges should read about 22M; commit that artifact as `bench/baseline.json` with a `note` pointing at the attribution note, so the recorded remainder over v0.3.6 is an explained cost rather than a blessing.


zonuexe:
Merged `origin/master` (#810, #813, #814) into the branch: against the new master (36,257,110 allocations, `make check` warning-free) the branch reads 21,835,993 (−39.8%), `--format json` byte-identical to that master, `make verify` green. The PR body's table is relative to the branch point `ffb456b0` and stands as the per-lever record. Left Draft per the #814 rule: `gh pr ready` and the merge wait for an APPROVE on GitHub.

PR #891, opened 2026-09-09
Publish the cache schema marker by rename

### LINKED ISSUE #807: Cache store: repair_writable_marker! races a concurrent constructor and can clear the root under a reader
> *This was generated by AI during triage.*

Found by the #799 lane (#804): `spec/rigor/cache/store_spec.rb:628` ("does not collide across concurrent writers targeting the same entry") failed once on CI shard 1 with `Errno::ENOENT` inside `Store#read_entry` (`File.binread(path)`), passed on re-run, and was clean across 25 local repetitions. It is unrelated to #804's change (the annotation-location Marshal patch and `FORMAT_VERSION` bump touch neither the store's write path nor this example).

The example spins 16 threads that EACH construct `Store.new(root: cache_root)` on a fresh root and race `fetch_or_compute` for the same entry. Constructing a store runs `repair_writable_marker!`: one thread can `File.read` `schema_version.txt` while another is mid-`File.write`, read a torn or empty value, conclude the marker mismatches, and `clear_cache_root!` out from under a sibling between that sibling's `File.file?` check and its `binread`. The cache-root clear is not guarded against a concurrent marker write, so the race is in the store's constructor, not in the atomic-replace path the example means to exercise.

Fix shape: either write the marker atomically (temp file + rename, the same discipline `atomically_replace` already has) so a reader never sees a torn value, or make the example construct ONE store (or pre-create the marker) before spawning the writers so it exercises only the entry-level race it names. The first is the real fix — a user running two `rigor check` processes over one fresh `.rigor/cache` hits the same window.

Pointers: `lib/rigor/cache/store.rb` (`repair_writable_marker!`, `clear_cache_root!`, `read_entry`, `atomically_replace`), `spec/rigor/cache/store_spec.rb:628`.

zonuexe:
Fixes #807

`Store#repair_writable_marker!` runs in the constructor and wrote `schema_version.txt` in place. Its reader is another process's constructor, which takes no lock, so `O_TRUNC`-then-write exposed a window in which the marker exists and is empty. A reader landing there concludes the marker disagrees and calls `clear_cache_root!` — under, among others, a sibling that is between its own `File.file?` and its `File.binread`, which is the `Errno::ENOENT` out of `#read_entry` the issue reports. Two `rigor check` processes over one fresh `.rigor/cache` are enough for it.

## What changed

- **The marker is published by rename.** The temp-file-then-rename half of `#atomically_replace` is extracted as `#publish_by_rename` so there is one implementation, not two. The marker uses that half alone: it needs the atomicity, but not the ordering a destination lock buys, since every writer of the marker writes the same value.
- **An EMPTY marker is treated as an ABSENT one** — rewritten, never cleared on. Emptiness carries no information: it is what a torn in-place write by an older Rigor looks like, what a crash between the create and the write leaves behind, and what `clear_cache_root!` itself passes through on its way to rewriting the marker. Destroying a whole cache root is the wrong answer to an ambiguous signal, and declining grants the entries below no trust they did not already have — the writable-store rules already keep and serve entries under a marker that is simply missing.
- **`#read_entry` tolerates an `ENOENT` between the existence check and the open**, as defence in depth, reading it as the same miss the existence check already reports for a file that was never there. The rescue is deliberately narrow: every other `SystemCallError` still raises, so a permission or I/O error stays visible instead of degrading into an endless recompute, and a present-but-corrupt file is untouched — that still goes through the envelope and payload checks.
- `docs/internal-spec/cache.md` records all three (the writable-store marker rules, the read fault-tolerance list, and the atomicity section).

## Pre-fix repro

Measured, not assumed.

| Probe | Pre-fix | Post-fix |
| --- | --- | --- |
| 12 processes constructing a Store over one FRESH root, 40 rounds (spurious `clear_cache_root!` calls) | 6 clears, in 4/40 rounds | 0 clears, 0/40 rounds |
| New example "survives many stores constructed over one root left stale by an earlier release" | red on 12 of 13 runs, `Errno::ENOENT … .entry` | green 10/10 |
| The three deterministic new examples (rename publication, empty marker, vanished entry) | red every run | green |

The reported CI failure was `Errno::ENOENT` inside `File.binread`; the concurrency example reproduces exactly that signature and backtrace frame.

## Not weakened

`spec/rigor/cache/store_spec.rb`'s existing "does not collide across concurrent writers targeting the same entry" is untouched — still 16 threads, each constructing its own `Store`. Two neighbouring examples were adjusted for the new write path rather than the assertion they make: the `SecureRandom.hex(4)` example now says `at_least(:once)` (one fetch over a fresh root draws two suffixes now that the marker uses the same helper), and the failing-rename example warms the marker before stubbing `File.rename` so the stub still lands on the ENTRY write it names.

`make verify` green; 1218 files inspected, no offenses; no `warning:` in the `check` output.

### DIFF
--- a/changelog.d/fixed/fix-807-marker-write-race.md
+++ b/changelog.d/fixed/fix-807-marker-write-race.md
@@ -1,0 +1,1 @@
+- **[cache]** Two `rigor check` processes sharing one `.rigor/cache` no longer race each other's schema-version marker, so neither clears the cache root out from under the other or fails with `Errno::ENOENT` ([#891](https://github.com/rigortype/rigor/pull/891)).
--- a/docs/internal-spec/cache.md
+++ b/docs/internal-spec/cache.md
@@ -431,7 +431,12 @@ A read encountering any of the following silently returns a
 cache miss; the producer block reruns and the next write
 overwrites the bad entry:
 
-- Missing entry file.
+- Missing entry file — including one that disappears between
+  the existence check and the open (`Errno::ENOENT`), which is
+  what a concurrent process clearing or compacting the root
+  looks like. Every other `SystemCallError` (permission, I/O)
+  still raises: a broken filesystem must stay visible rather
+  than degrade into an endless recompute.
 - Entry shorter than the minimum envelope (header + trailer).
 - Mismatched magic + format-version header.
 - Mismatched trailing SHA-256.
@@ -462,6 +467,15 @@ with different semantics for a writable vs. a read-only store:
 **Writable store:**
 
 - Marker missing → write the current value, proceed. Disk available.
+- Marker present but EMPTY → treated exactly as missing: write the
+  current value, proceed, keep the entries. Disk available.
+  Emptiness carries no information — it is what a torn in-place
+  write looks like, what a crash between the create and the write
+  leaves behind, and what the clear-then-rewrite below itself passes
+  through — and destroying a cache root is the wrong answer to an
+  ambiguous signal. Declining grants the entries no trust they did
+  not already have, since a missing marker over existing entries is
+  kept and served by the rule above.
 - Marker matches → proceed. Disk available.
 - Marker disagrees → wipe every entry under `<root>` (`unlink` every
   child via `FileUtils.rm_rf`), rewrite the marker, and proceed as if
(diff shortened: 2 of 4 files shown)


zonuexe:
The cross-process probe is preserved on branch `probe-807-marker-race` (`tmp/probe-807/`, with a README): `race_probe5.rb` is the fork harness that reported 6 spurious cache-root clears across 4 of 40 rounds pre-fix and 0 of 40 after. The in-process variants are kept alongside it as the record of what did NOT reproduce.

Issue #424, opened 2026-08-19
Reconcile the effect-collection cost measurement, then optimise if needed


zonuexe:
## Parent

#376 (ADR-103 umbrella). Split out of #409's WD13 checkbox, which is now measured and **fails**.

## The gap

ADR-103 § WD13 budgets effect collection at *"≤ ~5 % wall / RSS on mastodon"*. Measured 2026-08-19 ([note](https://github.com/rigortype/rigor/blob/master/docs/notes/20260819-wd13-effect-budget-verification.md), [evidence on #409](https://github.com/rigortype/rigor/issues/409#issuecomment-5336151269)):

| target | files | wall off → on | RSS |
| --- | --- | --- | --- |
| redmine | 347 | 8.69 → 9.53 s (**+9.7 %**) | +7 % |
| mastodon | 1,312 | 14.57 → 18.63 s (**+27.9 %**) | noise |

Roughly 2× the bound on redmine and 6× on mastodon. RSS is fine; wall time is the problem. Both arms ran sequentially (`parallel.workers` defaults to `0`), which is the default user experience and the population default-on would newly charge.

This issue is the **v0.3.5 → v0.3.9 optimisation window** the owner opened. v0.4.0 is a Go / No-Go on default-on decided against the CI band, and the bound is not being widened to fit the measurement.

## First lead, unverified

`Effects::Scanner` is explicitly *"a separate walk … taken here"* rather than riding `ScopeIndexer`'s existing `def` descent — ADR-103 WD13 preferred the shared descent and the separate walk was taken because the scanner must also attribute each recorded call node to its enclosing unit, and doing that inside the indexer would put an effects-shaped concern on the hot path for every run.

That trade bought a byte-identical `rigor check` when effects are **off**, which is not negotiable and must survive any change here. But it means an effects-**on** run walks every file's tree twice. A per-file cost proportional to the tree is the right order of magnitude for what was measured.

**Confirm or refute this before optimising anything.** A profile (stackprof in a throwaway `GEM_HOME`, cwd = target, per the corpus perf notes' method) attributing the delta is the first deliverable, not a rewrite.

## Candidate directions, in rough order of expected yield

Each needs the profile first; none is committed to.

1. **Ride `ScopeIndexer`'s descent after all**, paying the attribution cost behind the same activation-count flag `Collector` already uses so the off path stays a plain integer read.
[...]

Issue #409, opened 2026-08-17
Effects default-on at v0.4.0 — graduation preconditions (ADR-103 WD15)


zonuexe:
## Parent

#376 (ADR-103 umbrella). Owner ruling 2026-08-17: effect collection and `effects.check` become **default-on at v0.4.0**, opt-out via `effects: false`; the bleeding-edge feature `effects-on-by-default` (#408) is the preview. ADR-103 WD15 records the preconditions; this issue tracks clearing them before the default flips.

## What to clear

- [x] **Non-fork backends carry the effect side-table** — **replaced, not cleared (WD16).** There is no thread backend and the Ractor one is dead upstream, so no work Rigor owns can clear this as written. Without `fork` a collecting run degrades to sequential and says so, and that degrade is sound: the sequential path collects identically, so Windows pays wall-clock rather than correctness. #410 and #414 closed.
- [x] **Vocabulary and `effect.*` ids settled** — **cleared (WD16).** Vocabulary 1 ships as it stands. The `mutate` spellings already agree with Steins (WD14); `io.db.*` is Rigor's own leaf set and evolution-safe; the application-meaning roots diverge architecturally rather than lexically (Steins holds ecosystem labels outside its builtin set) and Rigor keeps them as builtin rows. Nothing renames, so nothing here gates the flip — a later Steins-driven spelling lands through `retired:` plus a vocabulary bump. Evidence on #378.
- [x] **WD13 budget re-verified** — **arbiter settled (WD16).** The advisory CI `effect-budget` job's band decides the mastodon half; a go/no-go taken on a developer host is what produced the retracted 2026-08-19 finding. The gitlab half (`Propagator.propagate` at 1.34 s against a 1 s bound) is unretracted and becomes #424's optimisation target across v0.3.5-v0.3.9; it does not gate the flip, provided the release notes disclose it.
- [x] **`effects.lsp` semantics** — **decided (WD16).** Editor mode stays effect-free, spelled as a real `effects.lsp` key defaulting to `false`, so the seat is not silently empty once the default flips. The key, its documentation and the guard land with the flip commit; a hover that reports a method's labels is a v0.4.x slice with its own consumer.
- [ ] **Migration note** in the release notes for projects carrying `%a{pure}` (their tags start being checked), and retirement of `effect.annotations-unchecked`.
- [x] **Snapshot taint-only rows** decided — **closed by #411, shipped in #415.** A row carrying only its taint is omitted by default; `--full` keeps it.
[...]

PR #969, opened 2026-09-09
Profile effect collection's wall-time delta and report the bound unreachable

zonuexe:
## What this is

#424's first deliverable, and only that: a profile attributing effect collection's wall-time delta, with a verdict on the double-AST-walk hypothesis. Note: [`docs/notes/20260910-effect-collection-profile.md`](https://github.com/rigortype/rigor/blob/fix-424-effect-collection-cost/docs/notes/20260910-effect-collection-profile.md). **No code change**, deliberately — the profile says the bound is not reachable by any bounded change to the scanner.

## Reproduced, on the configuration that matters

`tool/effect_budget.rb`, 3 interleaved reps, redmine (347 files), **no plugins**, `--no-cache`, sequential, YJIT on: wall **9.05 s → 10.07 s, +11.3 %**, arms cleanly separated (off max 9.18 < on min 9.93). RSS +8.6 % with overlapping ranges, so no finding there.

That agrees with the CI `effect-budget` job's +10.9 % on plugin-less mastodon, and it also reconciles the contradiction the 2026-08-19 note left open: the +3.4 % figures loaded the full Rails plugin list, and a larger base divides the same per-file linear cost down. The plugin-less number is the one WD13 has to hold for.

## Attribution (stackprof, `:wall`, 1 ms, both arms, two rep pairs)

Shares are of the delta (710 and 793 samples).

| frame (inclusive) | share of delta |
| --- | --- |
| `Effects::Scanner.scan` — the second walk | **~36 %** |
| — of which `UnitScan#run` (method bodies) | ~22 % |
| — of which the `def` identity descent | ~12 % |
| `Collector.record_call` + `record_unresolved` + `active?` | ~15 % |
| `Propagator.propagate` | ~11 % |
| GC delta | ~9 % |
| dispersed remainder | ~29 % |

## Verdict: confirmed in kind, refuted in size

The scanner really does walk every file a second time, and it is the largest single line. But at ~36 % of a +11.3 % delta, removing it *entirely* still lands near +7 %.

The sub-split is what decides the issue's three candidate directions, and it refutes all of them:

- **Direction 1 (ride `ScopeIndexer`'s descent).** The descent is the smaller half. `UnitScan#run` costs the same whichever parent reaches it, so the recoverable part is ~12 % of the delta — best case ~1.3 pp off 11.3 pp, bought by putting an effects-shaped concern on every run's hot path. That is the trade WD13 declined when the upside was assumed to be larger.
- **Direction 2 (drive the walk off the recorded call-site table).** The table covers call sites; `UnitScan` also needs ivar writes, globals and local ownership, so the body still has to be visited.
- **Direction 3 (skip units that cannot contribute).** Deciding a unit is empty requires the body visit being avoided.

The two remaining large lines are not walks: per-dispatch recording (~15 %, already ADR-103 WD13's minimal shape) and the GC delta (~9 %, the allocation collection itself makes). Neither moves without changing what collection proves.

**So: the ≤ 5 % bound is not reachable by a bounded change to the scanner.** Reaching it needs either one shared descent that subsumes `UnitScan`'s body work as well as the `def` identity walk — a redesign paid by every run, effects on or off — or a reduction in what collection proves. Per #424's own last acceptance criterion, that is a No-Go input for #409 rather than something to trim quietly.

## One trap worth flagging

`EnvelopeIndex.build` reads as a third of the delta on inclusive samples. It is not a contributor: its whole subtree is the RBS environment build, which costs 211 in the on arm against 210 in the off arm. The index only asks for the environment earlier than anything else does. Profiling the on arm alone would have produced a confident wrong answer here.

## Not done, and why

No optimisation, per the above. No changelog fragment: nothing user-facing changed. The off path is untouched, so `rigor check` with no `effects:` block is byte-identical by construction rather than by assertion.

`Propagator.propagate` — the live target ADR-103 WD16 re-points this issue at — is 0.08 s at redmine scale. The bound it answers to is the ≤ 1 s one at gitlab scale, which this measurement does not reach.

Refs #424 — deliberately not `Fixes`: this settles the per-project half (double-walk hypothesis, and the verdict that the bound is unreachable without changing what collection proves), but ADR-103 WD16's live target for this issue is `Propagator.propagate` at gitlab scale, which is untouched here.

### DIFF
--- a/docs/notes/20260910-effect-collection-profile.md
+++ b/docs/notes/20260910-effect-collection-profile.md
@@ -1,0 +1,111 @@
+# Where effect collection's wall-time delta actually goes (#424)
+
+Status: measurement note. **No code change**, and this note argues that the bounded changes #424 lists
+cannot reach WD13's budget. The double-AST-walk hypothesis is **confirmed as the largest single
+contributor and refuted as a sufficient explanation** — it is roughly a third of the delta, and the
+part of it a shared descent could recover is roughly a seventh.
+
+[ADR-103](../adr/103-effect-labels.md) § WD13 budgets collection at "≤ ~5 % wall / RSS"; § WD16
+(2026-08-22) makes the CI `effect-budget` job the arbiter for the mastodon half and re-points the issue
+at the `Propagator.propagate` bound. This note profiles the per-project cost anyway, because #424's
+second comment asks for exactly that on the **plugin-less** configuration — the shape CI measures and
+the population a default-on flip would newly charge.
+
+## The A/B, reproduced locally on the population that matters
+
+`tool/effect_budget.rb`, 3 interleaved reps, redmine (347 files, `app` + `lib`), **no plugins**,
+`--no-cache`, sequential (`parallel.workers` default `0`), YJIT on:
+
+| metric | off median | on median | Δ | arms separated? |
+| --- | --- | --- | --- | --- |
+| wall | 9.05 s (9.02–9.18) | 10.07 s (9.93–10.14) | **+11.3 %** | yes |
+| peak RSS | 302.4 MB (294.8–323.5) | 328.3 MB (323.0–332.1) | +8.6 % | no |
+
+The wall arms separate cleanly, and +11.3 % on redmine-without-plugins agrees with the CI job's
++10.9 % on mastodon-without-plugins. It also reconciles the contradiction the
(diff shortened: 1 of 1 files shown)


zonuexe:
Landed on master directly as a Markdown-only change (repo rule: docs commit straight to master); the note is now at docs/notes/20260910-effect-collection-profile.md.

PR #983, opened 2026-09-11
Bump up version to 0.3.9

zonuexe:
v0.3.9 is the last release of the 0.3 line and the largest fix batch of it: over a hundred pull requests closing false positives across inference (block returns, in-place mutation, constant and header resolution, implicit-self calls), plus the cache and incremental-mode staleness bugs where a warm run kept serving an answer the tree no longer held. Ranges now read as Ruby spells them (`Integer[1..10]`, `Float[0.0..1.0]`, ADR-109) and `int<min, max>` is deprecated ahead of v0.4.0. It also ships the capability-role catalog, `--fail-on`, the manifest-driven plugin-gap advisory, a 40% allocation cut on `rigor check`, and exit 70 on a fatal analysis error.

Sealed section: 91 `changelog.d/` fragments consolidated into `## [0.3.9]`; every merged PR since v0.3.8 is either linked from an entry or internal (docs, specs, CI plumbing). `make verify`, `make docs-check`, `git diff --check` and `gem build` are green locally.

Note for users: the first run after upgrading re-analyses once, because both cache schemas moved (`Cache::Descriptor::SCHEMA_VERSION` 9, `IncrementalSnapshot::SCHEMA` 22).


zonuexe:
Re-cut from master after #985 (the Mastodon crash) and the perf-baseline recalibration; superseded by the new release PR.

Issue #992, opened 2026-09-12
Arity-check source-defined methods that carry no declaration: the DiscoveryIndex records no parameter shape


zonuexe:
## What to build

After #991, a method the project defines in Ruby source and does **not** declare anywhere is still exempt from `call.wrong-arity`. The reason is structural: `Scope#discovered_method?` answers from a `DiscoveryIndex` table that maps `method_name => :instance | :singleton`, with no parameter shape in it, so there is nothing for the rule to check against.

This slice gives the index that shape and lets it be the authority when no RBS declaration exists. Record, per discovered method, the `Prism::DefNode` parameter counts the arity envelope needs — required, optional, rest, post, keyword (required / optional / rest), and block — then have `wrong_arity_diagnostic` fall back to that envelope instead of declining outright.

`Foo.new.f` against `def f(num)` is the target case: an arity mistake that is plainly visible in the same file, on a method nobody annotated, in a project with no `sig/` at all. That is the shape most adopting projects are in on day one, so this is where the rule earns most of its value — and it is also where a false positive is most expensive (ADR-5, and the standing "false positives outrank worst-case static reading" rule in `AGENTS.md`).

The false-positive envelope is the work, not the plumbing. At minimum the fallback must decline when:

- the method is recorded from `define_method` rather than a `DefNode`, where no static parameter list exists
- the receiver's class (or any ancestor reachable from it) defines `method_missing` / `respond_to_missing?`
- the name is aliased, or the class is reopened with a second `def` of the same name and a different shape — the index must not silently keep one of them
- a project `prepend`/`include` can supply another definition of the name, or an ADR-17 `pre_eval:` monkey-patch does
- the receiver shape is one the sibling rules already refuse to enumerate (`unenumerable_receiver_shape?`, #879) or is an ADR-67 WD6b inferred-parameter receiver
- a plugin contributes to the receiver's method surface

`super`, `yield`, and blocks passed with `&` do not change a `def`'s own positional envelope and should stay in scope.

Landing behind a configuration gate, default off for one release, is acceptable if the corpus evidence is not conclusive — state which way it landed and why.

## Repro

```ruby
class A
  def f(num)
    num
  end
end

A.new.f       # expected: call.wrong-arity (given 0, expected 1) — silent today
A.new.f(1, 2) # expected: call.wrong-arity (given 2, expected 1) — silent today
```

[...]

Issue #917, opened 2026-09-09
`call.wrong-arity` on correct code: a class whose RBS declares no `initialize` inherits `BasicObject`’s `() -> void`, so `.new` is checked as arity 0


zonuexe:
A class the loaded RBS declares but gives no `initialize` takes its `.new` envelope from `BasicObject#initialize`, which is nullary — so every argument at a real constructor is reported.

```ruby
require "rubygems"
Gem::Specification.new("mygem", "1.0.0")
```

Actual (v0.3.8 `ffb456b0` and master `5caf61a4`):

```
3:20 error: wrong number of arguments to `new' on Gem::Specification (given 2, expected 0)
```

The code is correct — it runs and builds the object:

```
$ ruby -e 'require "rubygems"; s = Gem::Specification.new("mygem", "1.0.0"); puts "#{s.name} #{s.version}"'
mygem 1.0.0
```

## Mechanism

Asked through the analyzer's own environment (`Rigor::Environment.for_project`), the resolution is entirely from `BasicObject`:

```
.new defined_in : ::BasicObject
.new types      : () -> ::Gem::Specification
#initialize     : ::BasicObject
```

`data/vendored_gem_sigs/rubygems/` declares plenty of `Gem::Specification`'s surface but no `initialize`, so the definition builder falls back to `BasicObject#initialize: () -> void`, `.new` derives `0..0`, and the arity rule treats that as a *known* envelope rather than as an absent declaration. The distinction the rule needs is between "this class declares a nullary constructor" and "no loaded RBS says anything about this constructor" — only the first is evidence.

The same shape reaches any class in that position; `Gem::Specification` is simply the one a corpus hits (7 rows across dependabot-core and gitlab-foss in my differential, all of them diagnostics that are wrong about the program).

## Why it is worth deciding rather than leaving

It is the failure direction the project's own zero-false-positive discipline exists for, and it is silent about its own weakness: an absent `initialize` is indistinguishable, at the rule, from a declared nullary one. Two ways out, both yours to choose between — decline the arity verdict when `.new`'s only `defined_in` is `BasicObject` and the receiver's own class declares no `initialize`, or supply the missing `initialize` declarations in the vendored gem sigs (which fixes `Gem::Specification` and not the general case).

Not affected, measured the same way: `Tempfile.new("prefix")` and `StringIO.new("body")` are silent — those classes declare their own constructors.

[...]

Issue #879, opened 2026-09-09
`call.wrong-arity` still enumerates a receiver `call.undefined-method` declines as unenumerable (the evidence #742 asked for)


zonuexe:
#739 / #742 stopped `call.undefined-method` speaking about a receiver whose method surface cannot be enumerated — a mixin-module type, a `Class` / `Module` value. `unenumerable_receiver?` has one caller, and `call.wrong-arity` keeps reading the narrower `unbounded_receiver_surface?`, so it still decides arity against the surface the other rule just declined.

`check_rules.rb`'s own comment on `unenumerable_receiver?` says why it was left that way:

> Kept beside `unbounded_receiver_surface?` rather than folded into it: that predicate has eight other callers (arity, the `raise` verdicts), and widening all of them is a separate change with its own evidence to gather.

This is that evidence, such as it is.

```ruby
require "digest"

def unenumerable_module_receiver(v)
  return unless v.is_a?(Digest::Instance)
  v.hexdigest(1, 2, 3)      # arity decided against the MODULE's declaration
end

def unenumerable_module_receiver_undefined(v)
  return unless v.is_a?(Digest::Instance)
  v.no_such_method_zzz      # #739: declined, the surface is not enumerable
end

def enumerable_control(v)
  return unless v.is_a?(String)
  v.upcase(1, 2, 3)         # control: a real class — must keep reporting
end
```

Actual (v0.3.8 `ffb456b0` and master `5f394719`):

```
6:5  error: wrong number of arguments to `hexdigest' on Digest::Instance (given 3, expected 0..1)
26:5 error: wrong number of arguments to `upcase' on String (given 3, expected 0..2)
```

The middle method is silent, as #739 intends. So on one receiver, in one run, the analyzer both declines to say which methods exist and asserts how many arguments one of them takes. The unsoundness is the same shape: `v` is an instance of some class that includes the module, and that class may define `hexdigest` with a different arity, in which case the call is correct code and this is a false positive.

## What I could NOT show

I tried to build a project-shaped repro — a project module with a method, an includer that overrides it with a wider arity — and it produces nothing at all, because the arity rule needs an RBS-declared signature and a project module's own `def` is not one:

```ruby
module Renderable
  def render(target) = target.to_s
end
class HtmlPage
  include Renderable
  def render(target, layout, locals = {}) = [target, layout, locals].to_s
end
def draw(v)
  return unless v.is_a?(Renderable)
[...]

PR #999, opened 2026-09-12
Check arity against a source-defined method's own signature

### LINKED ISSUE #991: call.wrong-arity exempts every source-defined method, including ones with a trustworthy signature
## What to build

`call.wrong-arity` never fires on a method the project defines in Ruby source, even when that method carries a complete, trustworthy signature. `CheckRules#wrong_arity_diagnostic` bails on `scope.discovered_method?(class_name, call_node.name, kind)` before it ever asks for the declaration, so a `def` in the project buys the method a blanket exemption from arity checking.

The exemption was written for methods that have no declaration at all: `Scope#discovered_method?` reads the `DiscoveryIndex` table, which maps `method_name => :instance | :singleton` and records **no parameter shape**, so at that point the rule has no arity to check against. That reasoning does not extend to a method the project also declares — in `sig/`, or inline via `# @rbs` / `#:` (ADR-93), which synthesizes a real RBS member.

`call.argument-type-mismatch` already made the opposite call on the same question, and says so in a comment beside its own gate: when the user supplies BOTH a `def` and an RBS sig, the sig is the authoritative parameter contract and calls are validated against it. The two rules read the same `trustworthy_signature` and should answer the same way about whose contract binds. Today they disagree, and the arity half is the silent one.

This slice narrows the exemption: source discovery suppresses the arity check only when no trustworthy signature is available for that receiver and method. When one is, the declared arity decides, under every guard the rule already applies (`unenumerable_receiver_shape?` per #879, `undeclared_constructor?` per #917, `anonymous_struct_new_call?`, and ADR-67 WD6b's inferred-parameter receiver decline).

Deriving arity from the `def` itself, for methods with no declaration at all, is deliberately NOT in this slice — that needs a parameter shape in the `DiscoveryIndex` and a much larger false-positive envelope.

## Repro

```ruby
class Foo
  # @rbs num: Float
  # @rbs return: Float
  def f(num)
    num
  end
end

Foo.new.f          # expected: call.wrong-arity (given 0, expected 1) — silent today
Foo.new.f(1.0, 2.0) # expected: call.wrong-arity (given 2, expected 1) — silent today
Foo.new.f("x")     # call.argument-type-mismatch DOES fire today, on the same declaration
```

Measured at `568138c2`. The third line proves the declaration is present, readable, and already authoritative for the sibling rule; only the arity rule declines to look at it.

## Acceptance criteria

- [ ] A call whose arity contradicts a `sig/`-declared signature reports `call.wrong-arity`, even when the project also defines the method in Ruby source
- [ ] The same holds for an inline `# @rbs` / `#:` declaration (ADR-93), whose synthesized member is what the repro above exercises
- [ ] A call to a source-defined method with NO declaration stays silent — unchanged from today
- [ ] The receiver-shape declines of #879 and #917 still hold: no new firing on an unenumerable receiver surface, and none on a class whose RBS declares no `initialize`
- [ ] `make verify` is green, and `make check` / `make check-plugins` stay clean — the change opens a rule over the whole self-check corpus, so a new firing there is either a real bug in `lib/` or an envelope that is still too wide
- [ ] A corpus pass over the survey projects shows no false positive attributable to this slice

## Blocked by

None - can start immediately.

zonuexe:
Closes #991

## What changed
`CheckRules#wrong_arity_diagnostic` bailed on `scope.discovered_method?(class_name, call_node.name, kind)` before it ever asked for a declaration, so a project `def` bought the method a blanket exemption from arity checking — even when it also carried a `sig/` entry or an inline `# @rbs` / `#:` annotation (ADR-93) that synthesized a real RBS member.
That reasoning is correct only for a method the project declares nowhere else: `Scope#discovered_method?` reads the `DiscoveryIndex` table, which records `method_name => :instance | :singleton` with no parameter shape, so at that point the rule genuinely has no arity to check against.
`call.argument-type-mismatch` already made the opposite call on the same question, reading the same `trustworthy_signature` lookup as authoritative even when the project also supplies a `def`. The fix removes the premature `discovered_method?` exemption from `wrong_arity_diagnostic`; the existing `method_def.nil?` guard (fed by `trustworthy_signature`) now does the same job precisely — silent when no trustworthy signature resolves anywhere in the receiver's ancestry, firing when one does, regardless of whether the project also defines the method in source.
Every guard the rule already applies stays in place and in the same order: `unenumerable_receiver_shape?` (#879), `undeclared_constructor?` (#917), `anonymous_struct_new_call?`, and the ADR-67 WD6b inferred-parameter receiver decline.

## Docs
`docs/type-specification/diagnostic-policy.md`'s `call.wrong-arity` entry now states the same-terms-as-argument-type-mismatch rule explicitly, alongside the existing #917 clause.

## Testing
New spec: `spec/integration/arity_declared_source_method_spec.rb` — three arms:
- a `sig/`-declared method the project also defines in source (fires)
- an inline `# @rbs`-declared method the project also defines in source, via the bundled `rigor-rbs-inline` plugin (fires)
- the same call shapes against a method with no declaration at all (stays silent — unchanged, issue #992's territory)

Commands run (all inside `nix develop --command`):
- `bundle exec rspec spec/integration/arity_declared_source_method_spec.rb` — 3 examples, 0 failures
- `bundle exec rspec spec/rigor/analysis/check_rules/ spec/integration/` — 2603 examples, 0 failures
- `bundle exec rubocop lib/ spec/ plugins/ examples/ apps/` (via `make lint`) — 1252 files, no offenses
- `make check` — clean (only the pre-existing `rbs.coverage.missing-gem` info notice)
- `make check-plugins` — clean (same pre-existing info notice)
- `make docs-check` — 464 examples, 0 failures, 1 pending (an unrelated, expected skip: `references/ruby` submodule not checked out here)

`make verify` was intentionally NOT run per the task's instructions (parallel lanes on this machine OOM under a full gate); the orchestrator runs it separately.


zonuexe:
Orchestrator review: the shape of the fix is right and the dangerous neighbour is safe — `class MyStr < String; def center(a, b, c); end; end` followed by `MyStr.new.center(1, 2, 3)` stays silent, so an inherited declaration is not being used to decide a call against the receiver’s own `def` (ADR-110 holds).

One acceptance criterion fails, though: a source-defined method with **no** declaration now fires, which is #992’s territory and carries none of #992’s false-positive envelope.

```
controls.rb:25:20: error: wrong number of arguments to `f' on DeclaredInline (given 0, expected 1) [call.wrong-arity]
controls.rb:26:20: error: wrong number of arguments to `f' on DeclaredInline (given 2, expected 1) [call.wrong-arity]
controls.rb:27:16: error: wrong number of arguments to `g' on Undeclared (given 0, expected 1) [call.wrong-arity]
```

Lines 25 and 26 are the intended fix (`DeclaredInline#f` carries `# @rbs num: Float` / `# @rbs return: Float`). Line 27 is `class Undeclared; def g(x); x; end; end` — no annotation, no `sig/` — and at the fork point `568138c2` none of the three fire, so line 27 is new.

Cause: under ADR-93 the bundled `rigor-rbs-inline` plugin synthesizes an RBS member for every `def`, annotated or not, so `trustworthy_signature` resolves something even for a bare `def` and `method_def.nil?` does not separate "the project declared this" from "the project merely defined it". The integration spec’s no-declaration arm passes only because that fixture project has no inline plugin contributing; a normal project does, by default.

The fix needs provenance rather than presence — the plugin’s `default_type` marker (`plugins/rigor-rbs-inline/lib/rigor/plugin/rbs_inline.rb`) exists to tell a defaulted type from an authored one. Picked up in a follow-up commit on this branch; staying Draft.

zonuexe:
Orchestrator review of `f7527916`: **`make verify` green** (exit 0), and the case the issue was filed for now reports.

```
foo.rb:18:13: error: wrong number of arguments to `f' on Foo (given 0, expected 1) [call.wrong-arity]
```

That is the playground snippet from the original report — `# @rbs num: Float` with no return annotation, `p Foo.new.f`. Silent at `568138c2` and at `6b53e12b`, reported now. The `rigor:v1:inferred-signature` split is the right shape: a parameter the author annotated is an assertion about the parameter list, and arity is a question about the parameter list.

Controls hold **cold**: the bare `def` and the mixin-reached bare `def` stay silent, both annotated-declaration calls fire, `sig/`-declared fires, authored `#: (String) -> untyped` fires.

One thing to know before this lands, and it is not this PR's bug: **the controls do not hold on a warm cache written by the parent build.** Populate `.rigor/cache` at `6b53e12b`, switch to `f7527916`, run warm, and the bare-`def` and mixin false positives come back — four diagnostics where both `--no-cache` and a cache-cleaned warm run give two. The cached synthesized RBS lacks the new annotation while the rule reading it is the new one, so the run is a mixture of two builds. The same experiment on PR #1006 (a `lib/`-only change) gives the new answer warm, so it is specific to the plugin-synthesis lane. Filed as #1009.

Practical consequence for anyone verifying this branch: use `--no-cache`, or clear `.rigor/cache` first. Staying Draft.

Issue #836, opened 2026-09-08
sig-gen treats a declared `void` return as a value type it can tighten


zonuexe:
Found by the `sig/` provenance audit (#825, PR #835): 7 of the 15 `tighter_return` proposals over Rigor's own `sig/` are methods declared `-> void` whose bodies happen to return a value. `RbsTypeTranslator` maps `void` to `Type::Top`, `Top.accepts` is total, so `compare_against_declared` reads every void mutator whose last expression is a value as a tightening candidate, and the provenance gate inherited that and demanded a "gap" marker for each.

This is a semantics error, not a false positive to tune. `void` is the author's contract that the return is not part of the API ("a value is returned, but don't rely on it" — a1-2 of the chibirigor appendix; `docs/type-specification/special-types.md` § `void`). A synthesizer that builds a type from the body always produces the type of the last expression; `void` never appears on the synthesis side and only works on the checking side. So a declared `void` is authored intent that no inference can produce — the same status ADR-107 gives parameter types — and sig-gen must never propose replacing it, `--overwrite` must never overwrite it, and the provenance gate must classify it as earned (`return_intent`), not as a recorded gap.

Fix: `compare_against_declared` classifies a declared `void` as `equivalent` regardless of the body; the auditor gains a `RETURN_INTENT` earned class; the seven `# sig-gen gap: #836` markers come out; ADR-14's contradiction rule records the exemption.

Related: ADR-107 (whose "return types are generated" rule gets the `void` exception in the same cycle), ADR-14, ADR-100, #825.

PR #1000, opened 2026-09-12
Propose the inferred return for a declared-untyped method

### LINKED ISSUE #995: sig-gen silently proposes nothing for a method with inline param annotations and an undeclared return
## What to build

A method whose parameters carry an inline `# @rbs` annotation but whose **return** is undeclared gets no `sig-gen` proposal, and no skip record either. The command prints `No candidates`, and `--format=json` returns `{"candidates": []}` — while the engine has in fact inferred the return and `rigor annotate` displays it on the `def` line.

That is the one case ADR-14 exists for: the author has said what they know and left the return to Rigor. `sig-gen` should propose it. Failing that, it must at least be *visible* — the JSON reports a skipped method with its `skip_reason` in every other decline, so this silent disappearance is worse than a skip.

The likely mechanism, to confirm before fixing: an inline `# @rbs num: Float` (ADR-93) synthesizes a real RBS member `def f: (Float) -> untyped`, so `lookup_existing_method` finds a declaration and the candidate takes the `compare_against_declared` path. There, `tighter?` answers false for a declared `untyped` — `untyped.accepts(inferred)` and `inferred.accepts(untyped)` are both `yes`, so the backward check fails it — and the candidate lands `Classification::EQUIVALENT`, which is neither emittable nor counted among the skips. The result is a declaration whose return says nothing being treated as a declaration the proposal cannot improve on.

A declared `untyped` return is not the author's statement of intent — that is what `void` is, and #836 already carved `void` out on exactly that reasoning (`declares_void?` returns `equivalent` carrying the author's own word). `untyped` is the absence of a statement, and it should behave like the no-declaration path: propose the inferred return as a tightening.

Scope note: whether the proposal writes to `sig/` or is rendered as the inline annotation the author is already using is a real question — inline and `sig/` are both ingested, and `sig/` wins on conflict. `--print` / `--diff` correctness comes first; pick one for `--write` and say why in the issue thread.

## Repro

```ruby
class Foo
  # @rbs num: Float
  def f(num)
    if (num > 0)
      [num, num.to_s]
    else
      [num, num.to_s]
    end
  end
end
```

At `568138c2`, in a project with a `.rigor.yml` and this file at `lib/foo.rb`:

- `rigor sig-gen --print lib/foo.rb` → `No candidates`
- `rigor sig-gen --print --format=json lib/foo.rb` → `{"candidates": []}`
- `rigor annotate lib/foo.rb` → `def f(num)  #=> [Float, String]`

The control, with the `# @rbs num: Float` line removed and a second undeclared method added, behaves correctly on the same run: one `new_method` candidate with its `rbs` line, and one `skipped` entry carrying `sig.skipped.untyped-return`. So both halves of the machinery work; only the partially-annotated method falls between them.

## Acceptance criteria

- [ ] The repro yields a proposal for `f`'s return under `--print` and `--format=json`, or — if the ruling is that it must not be proposed — a `skipped` entry with a `skip_reason` naming that decision; silence is not an acceptable outcome either way
- [ ] A declared `untyped` return is handled like an absent declaration rather than like a contract, and `void` keeps its #836 treatment unchanged — a spec pins both
- [ ] Specs cover the inline-annotated (ADR-93) and `sig/`-declared `-> untyped` spellings, since both reach the same comparison
- [ ] No new proposal appears for a method whose declared return is a real type — the `equivalent` classification keeps its meaning
- [ ] `make verify` green
- [ ] A `sig-gen` pass over this repo's own `lib/` shows the delta is confined to methods whose declared return is `untyped`

## Blocked by

None - can start immediately.

zonuexe:
Closes #995.

## The bug

A method whose parameters carry an inline `# @rbs` annotation (ADR-93) but whose return is undeclared got no `sig-gen` proposal, and no skip record either: `--print` said `No candidates`, `--format=json` returned `{"candidates": []}`, and `rigor annotate` showed the return was inferred all along.

## Root cause (confirmed)

The inline `# @rbs num: Float` synthesizes a real RBS member `def f: (Float) -> untyped`, so `lookup_existing_method` finds a declaration and the candidate takes the `compare_against_declared` path. There, `tighter?`'s backward acceptance check answers "yes" for a declared `untyped` in both directions — `untyped` accepts and is accepted by everything — so the candidate landed `Classification::EQUIVALENT`: not emittable, and not a `skipped` row either. It vanished from both `--print` and the JSON payload with nothing to explain why. The same comparison is reachable with a hand-written `sig/`-declared `-> untyped` return, not only the ADR-93 synthesized spelling.

## The fix

A declared `untyped` return says nothing at all — the absence of a statement — unlike a declared `void`, which is the author's own word that the value is not part of the contract (#836, `declares_void?`). `compare_against_declared` now checks `declared_untyped?` right after the existing `declares_void?` check (so `void` is unaffected — it is still checked first and still short-circuits) and, when the declared return is `Type::Dynamic` (RBS `Bases::Any`), builds the same `tighter-return` candidate a method with no declaration at all would get, carrying `"untyped"` as `declared_return_rbs` so `--diff` and the `[tighter, was: untyped]` print tag still tell the reader a declaration existed.

`lib/rigor/sig_gen/generator.rb`:
- `declared_untyped?(declared)` — true when the declared return translates to `Type::Dynamic`.
- `declared_untyped_candidate(...)` — built like `new_method_candidate`, not the ordinary tightening branch: a declared `untyped` carries no information to weigh `tighter?` or `literal_decline?` against, the same position a method with no declaration is in.

Docs: `docs/adr/14-rbs-sig-generation.md` gets a new dated paragraph in § "The inference-vs-RBS contradiction rule" (mirroring the `void` / RBS-literal paragraphs already there) plus a revision-history line.

Specs: `spec/rigor/sig_gen/generator_spec.rb` adds a `describe "#run when the declared return is \`untyped\`"` block covering both spellings that reach the comparison (a hand-written `sig/` `-> untyped` return, and the ADR-93 inline-annotated repro from the issue), plus two guard specs: `void` still classifies `equivalent` unchanged, and a real declared type with a matching inferred return still classifies `equivalent` (no new proposal).

## The `--write` open question

The issue asks this to be answered rather than guessed at silently. I ran `--write` against the repro (no `sig/` file yet, only the inline `# @rbs num: Float` annotation) and it creates a brand-new `sig/foo.rbs`:

```
class Foo
  def f: (untyped) -> [Float, String]
end
```

This is `writer.rb`'s existing, unmodified behavior for any `NEW_METHOD` / `TIGHTER_RETURN` candidate reaching a class with no existing `sig/` file — I did not touch `writer.rb`. It is a real tension worth flagging: the author's inline annotation already declares `num: Float`, but sig-gen's parameter policy always renders `untyped` for the parameter list (ADR-5 clause 2 — parameters are never auto-tightened), so the new `sig/` file's `(untyped)` parameter is *less* precise than the inline annotation sitting right next to it in the source, and per AGENTS.md's "where `sig/` declares the same member, `sig/` wins" rule, the new file's `untyped` param would now shadow the inline `Float` one at check time. Writing back into the inline annotation instead (rendering only the return clause the author left bare) would avoid fragmenting authorship across two files for one method, but that is a bigger `--write`-path change than this issue's scope — I left `--write` itself untouched and am recommending, not deciding, that a follow-up route `--write` back into an existing inline annotation's own return clause when that is where the declaration originated, rather than forking a parallel `sig/` file. Filing as a separate issue seems right rather than deciding it here.

## Verification

Inside the Nix Flake (`nix --extra-experimental-features 'nix-command flakes' develop --command <cmd>`), from the worktree root:

- `bundle exec rspec spec/rigor/sig_gen/generator_spec.rb` — 101 examples, 0 failures.
- `bundle exec rspec spec/rigor/sig_gen/ spec/rigor/cli_spec.rb -e "sig-gen"` — 29 examples, 0 failures.
- `bundle exec rspec spec/docs/` — 464 examples, 0 failures, 1 pending (unrelated: an optional `references/ruby` submodule not checked out here).
- `make lint` — clean (1251 files, 0 offenses; a first draft tripped `Lint/ConstantDefinitionInBlock` / `RSpec/LeakyConstantDeclaration` on a spec-file constant, fixed by using a helper method instead).
- `make check` — clean (exit 0; only the pre-existing `rbs.coverage.missing-gem` `:info`).
- `make check-plugins` — clean (exit 0; same pre-existing `:info`).

I did not run `make verify` per the orchestrator's instructions (parallel lanes on this machine).

## Confining the delta

Ran `sig-gen --print --format=json lib` over this repo's own `lib/` twice — once with this fix, once with master's `generator.rb` swapped in temporarily (same commit, `git show origin/master:... > generator.rb`, restored immediately after) — and diffed the two JSON payloads by `(file, class, method, kind)`:

- master: 5236 candidates, fixed: 5256 candidates.
- 20 added, 0 removed, 0 mutated (every candidate present in both runs has byte-identical content).
- All 20 added rows: `"classification": "tighter_return"` and `"declared_return_rbs": "untyped"` — exactly the shape this fix targets, nothing else moved.

### DIFF
--- a/changelog.d/fixed/sig-gen-declared-alias-return.md
+++ b/changelog.d/fixed/sig-gen-declared-alias-return.md
@@ -1,0 +1,1 @@
+- **[sig-gen]** A declared return spelled through a project's own RBS type alias (e.g. `-> Type::t`) is no longer misread as a bare `untyped`, which used to spuriously propose the inferred return as a `tighter-return` even when the alias already named it exactly. ([#1000](https://github.com/rigortype/rigor/pull/1000))
--- a/changelog.d/fixed/sig-gen-untyped-declared-return.md
+++ b/changelog.d/fixed/sig-gen-untyped-declared-return.md
@@ -1,0 +1,1 @@
+- **[sig-gen]** A method whose declared return is `untyped` (including one ADR-93 synthesizes from a parameter-only inline `# @rbs` annotation) now gets a `tighter-return` proposal for its inferred return instead of silently disappearing from `--print` and the JSON payload. ([#1000](https://github.com/rigortype/rigor/pull/1000))
--- a/docs/adr/14-rbs-sig-generation.md
+++ b/docs/adr/14-rbs-sig-generation.md
@@ -661,6 +661,22 @@ always answered this way when the declaration happened to be a UNION —
 refuses it — and whether the author wrote `String` or `String?` was never a
 decision about literals.
 
+**2026-09-12 — a declared `untyped` return is exempt from the rule in the OPPOSITE direction: it is treated
+like no declaration at all, and the inferred return IS proposed.** `void` and the RBS literal case above are
+both about a declaration that says MORE than the body proves — an author's abstraction the body cannot derive.
+A declared `untyped` (RBS `Bases::Any`) is the opposite: it says NOTHING. `tighter?`'s acceptance check cannot
+tell "nothing" apart from "something real the inferred form would narrow" — `untyped` accepts and is accepted
+by everything, so the backward half of the check, which exists to catch the second case, also caught the
+first, and `compare_against_declared` classified the method `equivalent` without ever emitting the return
+inference actually proved. Worse than a decline: `equivalent` is not in `Classification::EMITTABLE` and is not
+a `skipped` row either, so the candidate disappeared from both `--print` and the JSON payload with nothing to
+explain why ([#995](https://github.com/rigortype/rigor/issues/995)). ADR-93's inline `# @rbs param: Type`
+annotation (no return clause) synthesizes exactly this declared shape for every method it documents, so the
+silent drop landed on the parameter-annotated, return-bare method ADR-14 exists to help. `declared_untyped?`
+short-circuits `compare_against_declared` before `tighter?` runs at all, building the same `tighter-return`
+candidate a method with no declaration would get — carrying `"untyped"` as `declared_return_rbs` so `--diff`
+and the `[tighter, was: untyped]` print tag still tell the reader a declaration existed.
+
 Two boundaries. The rule is on the ERASURE, so a `Type::Constant` RBS has no
 literal spelling for still proposes its nominal: `def pi: () -> Numeric` against
 a body of `3.14` proposes `Float`. And it binds only against an EXISTING
@@ -700,3 +716,7 @@ override anywhere — has no wider declaration to find.
   same reading: the declared type is the author's abstraction over the body.
   Clause 3 of § "What 'more precise' means for tighter-return mode" and the
   second dated paragraph in § "The inference-vs-RBS contradiction rule".
+- 2026-09-12 — #995 exempted a declared `untyped` return, on the opposite
+  reading from `void` / the RBS-literal case: `untyped` says nothing, so it is
+  treated like no declaration at all and the inferred return is proposed. See
+  the third dated paragraph in § "The inference-vs-RBS contradiction rule".
--- a/lib/rigor/sig_gen/generator.rb
+++ b/lib/rigor/sig_gen/generator.rb
@@ -44,6 +44,11 @@ module Rigor
     # - A declared `void` return is never compared as a value type. It is the author's statement that the
     #   return is not part of the contract, which no inference can synthesize, so the method classifies
     #   `equivalent` carrying `void` itself as the declared spelling (#836; see {#declares_void?}).
+    # - A declared `untyped` return (the RBS `Bases::Any` spelling, including the one ADR-93 synthesizes for
+    #   a parameter-only inline `# @rbs` annotation) is the ABSENCE of a return statement, not the author's
+    #   word about it — the opposite of `void`. It is treated like no declaration at all: the inferred return
+    #   is proposed as `tighter-return`, carrying `untyped` as the declared spelling so `--diff` still shows a
+    #   declaration existed (#995; see {#declared_untyped?}).
     # - A proposal that erases to an RBS literal never tightens an existing declaration: the declared type is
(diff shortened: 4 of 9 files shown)


zonuexe:
Picked this back up to close the ADR-107 G3 gate failure `make verify` reported (green on the targeted specs, red on the full gate).

Root cause of all 22 unmarked tighter-returns: `Generator#translate_method_type_return` called `RbsTypeTranslator.translate` with no `alias_expander:`, so a declared return spelled through a project alias (`Type::t`, `Environment::ordering`, `Inference::closure_classification`) degraded to `Dynamic[Top]` exactly like a bare `untyped` would — which is exactly the shape `declared_untyped?` (this PR's own new check) treats as "propose a tightening." Most of the 22 were declarations that were already exactly right; they just couldn't be told apart from an absent declaration.

Four commits on top of the original two:

1. Pass the environment's `RbsLoader` as `alias_expander:` (the idiom `check_rules.rb` already uses). This alone resolves most of the 22 back to their correct classification and also surfaced one previously-hidden genuine tightening (`ScopeIndexer.finalize_constant_writes`) and fixed one unrelated pre-existing misclassification (`Reflection#constant_type_for`).
2. Apply the four remaining tightenings that are genuinely single-implementation, always-true facts (`Plugin::Base.node_rule`, `#diagnostic`, `#cache_for`, `ScopeIndexer.finalize_constant_writes`).
3. Mark the two remaining genuine engine gaps with filed issues: #1007 (`dynamic_return_type`'s real value flows through an `instance_exec`'d plugin block sig-gen can't see through) and #1008 (`return_summaries`'s Hash is populated by a sibling method sig-gen doesn't track through).
4. Lower the six `SIG_PROVENANCE_RESIDUE` pins to match, verified against a real gate run.

`make verify` is green (11112 examples, lint/check/check-plugins clean). Still Draft.

PR #1001, opened 2026-09-12
Project Integer#to_s / Float#to_s to numeric-string refinements

### LINKED ISSUE #993: Integer#to_s and Float#to_s land at bare String: project them to numeric-string, and to Float only with a finiteness proof
## What to build

`Integer#to_s` and `Float#to_s` on a receiver that is not a literal both land at bare `Nominal[String]`, even though Rigor already has every piece needed to say more: `Type::Combinator.numeric_string` / `.non_empty_string`, the `finite-float` refinement, and the `Float#finite?` truthy-edge narrowing that produces it (ADR-109 WD5).

Add the numeric-to-string projections to the shape-dispatch tier that already projects `Refined[String, *]` receivers (`REFINED_STRING_PROJECTIONS` in `MethodDispatcher::ShapeDispatch`), extending it to numeric receivers:

- **`Integer#to_s` with no argument** → `numeric-string`. Every Ruby Integer prints as a decimal integer literal, sign included, and `numeric-string` spans the signed grammar. `decimal-int-string` is the tighter true answer and is preferable if the carrier composes; `numeric-string` is the floor.
- **`Integer#to_s(base)` with a non-decimal base** → `non-empty-string` only. `255.to_s(16)` is `"ff"`, which is not a Ruby numeric literal (`0xff` would be).
- **`Float#to_s`** → `non-empty-string` by default, and `numeric-string` **only with a finiteness proof**: the receiver is `finite-float`, a `FloatRange` contained in `-Float::MAX..Float::MAX`, or a finite `Float` literal. A bare `Float` admits `Float::INFINITY`, `-Float::INFINITY` and `NAN`, which print as `"Infinity"`, `"-Infinity"` and `"NaN"` — all non-empty, none of them numeric literals.

The finiteness condition is the whole point of the issue, and it is the Ruby counterpart of the ruling recorded upstream in `rigortype/steins#663`, where PHPStan's unconditional `numeric-string` for a float's string conversion is refuted by `INF` / `NAN` with no diagnostic. Ruby's `Float#to_s` behaves the same way; do not claim numericness for a `Float` that has not been proven finite.

`Float[0.0..]` is specifically NOT a finiteness proof: per `docs/type-specification/imported-built-in-types.md`, `Float[0.0..]` includes `+Infinity` and only `Float[0.0...Float::INFINITY]` excludes it. The narrowing that `num > 0` produces is the inclusive form, so the common guarded shape lands on the non-finite side of this rule.

`#inspect` is the same value grammar as `#to_s` for both classes and should get the same treatment, or an explicit note saying why not.

## Repro

```ruby
class Probe
  # @rbs i: Integer
  # @rbs f: Float
  def probe(i, f)
    a = i.to_s          # today: String       — want: numeric-string
    b = f.to_s          # today: String       — want: non-empty-string
    c = i.to_s(16)      # today: String       — want: non-empty-string
    d = f.finite? ? f.to_s : "x" # today: "x" | String — want: "x" | numeric-string
  end
end
```

Measured at `568138c2`. Literal receivers already fold exactly (`1.1.to_s` is `"1.1"`, `Float::INFINITY.to_s` is `"Infinity"`, `3.to_s(2)` is `"11"`), and `f.finite?` already narrows its truthy edge to `finite-float` — the proof is available at the point the projection needs it.

## Acceptance criteria

- [ ] Each projection above has a spec asserting the produced carrier, including the negative cases: bare `Float` does NOT reach `numeric-string`, and neither does `Float[0.0..]`
- [ ] A spec pins the three non-finite spellings (`Float::INFINITY`, `-Float::INFINITY`, `Float::NAN`) as inhabitants that make the unconditional claim false, so a later widening cannot quietly re-introduce it
- [ ] `f.finite?` followed by `f.to_s` reaches `numeric-string` on the truthy edge and does not on the falsy edge
- [ ] `docs/type-specification/` records the projection table in the same commit as the behaviour
- [ ] No new diagnostic fires on the survey corpus attributable to the tighter result type — this is a precision change, not a diagnostic change

## Blocked by

None - can start immediately.

zonuexe:
## Summary

Closes #993.

`Integer#to_s` and `Float#to_s` on a non-literal receiver landed at bare `Nominal[String]`, even though `REFINED_STRING_PROJECTIONS`'s neighbouring tier already had `Type::Combinator.numeric_string` / `.non_empty_string` and the `finite-float` refinement available at exactly the point the projection runs.

## What changed

`lib/rigor/inference/method_dispatcher/shape_dispatch.rb`:

- `Nominal[Integer]#to_s` (no arg, or the literal base `10`) and `#inspect` now project to `decimal-int-string` — the predicate's leading `-?` already spans the signed grammar, so an unbounded `Integer` receiver needs no separate sign proof.
- `Nominal[Integer]#to_s(base)` for any other base — a different literal or a value that is not statically known — projects to `non-empty-string` only: `to_s(base)` never returns `""`, but a non-decimal base's digits (`255.to_s(16)` is `"ff"`) match no registered refinement.
- `Nominal[Float]#to_s` / `#inspect` project to `non-empty-string`: a bare `Float` has no finiteness proof, and `Infinity` / `-Infinity` / `NaN` all print as non-empty, non-numeric-literal strings.
- A new `FloatRange` receiver handler projects `#to_s` / `#inspect` to `numeric-string` exactly when both bounds are finite doubles (`min.finite? && canonical_max.finite?`, the same check `ConstantFolding`'s `float_range_finite` fold uses). This covers `finite-float`, any narrower bounded sub-range, and the `f.finite?` truthy-edge narrowing (ADR-109 WD5) that produces `finite-float` — while correctly excluding `Float[0.0..]`, whose endless upper bound is `+Infinity` and is therefore NOT a finiteness proof.

`docs/type-specification/imported-built-in-types.md` records the projection rule for both classes in the "Numeric refinement scope" section, in the same commit.

Two existing fixtures' golden precision snapshots changed as an expected side effect: `container_size` (`rand(100).to_s` is now `decimal-int-string` instead of `String`) and `reduce_symbol` (`rand.to_s` is now `non-empty-string`, so the `:+`-reduced result is too). Both are genuine precision improvements, reviewed and accepted.

## Test plan

- [x] `bundle exec rspec spec/rigor/inference/method_dispatcher/shape_dispatch_spec.rb spec/rigor/inference/method_dispatcher/constant_folding_spec.rb` — new unit specs for `Nominal[Integer]`, `Nominal[Float]`, and `FloatRange` `#to_s`/`#inspect`, including the negative cases (`Float[0.0..]`, `non-nan-float` do NOT reach `numeric-string`) and the three non-finite `Float` literal spellings pinned via `ConstantFolding`.
- [x] `bundle exec rspec spec/integration/type_construction_spec.rb spec/integration/precision_snapshot_spec.rb` — new integration fixture `numeric_to_s_refinements.rb`, plus `float_comparison_narrowing.rb` extended with the `f.finite?` truthy/falsy `.to_s` acceptance case.
- [x] `make lint` — clean.
- [x] `make check` / `make check-plugins` — clean (only the pre-existing `rbs.coverage.missing-gem` info notice), no diagnostic moved.


zonuexe:
Orchestrator review: **`make verify` green** on this branch (exit 0), `git diff --check` clean.

Verified the projections independently rather than from the specs: `Integer#to_s` → `decimal-int-string`, bare `Float#to_s` → `non-empty-string` (correctly **not** numeric-string), `Integer#to_s(16)` and a statically-unknown base → `non-empty-string`, and the finiteness path — `if f.finite?` then `f.to_s` → `numeric-string`, with the falsy edge staying `non-empty-string`. The two fixture/snapshot edits are genuine precision gains: `rand(100).to_s` is a `decimal-int-string`, and `rand.to_s` is non-empty because `to_s` is total.

Two things found while checking, neither a defect in this PR and neither blocking it:

- The same guard in its **ternary** spelling does not narrow at all — `f.finite? ? f.to_s : "x"` gives `"x" | non-empty-string` where the `if` form gives `numeric-string`. Confirmed identical at the fork point `568138c2`, so it predates this work: #1003.
- The existing `IntegerRange#to_s` tier maps base 16 / base 8 to `hex-int-string` / `octal-int-string`, which this PR deliberately left alone. Worth confirming separately whether those predicates require the `0x` / `0o` prefix — if they do, the mapping claims a refinement that `255.to_s(16) == "ff"` refutes.

Staying Draft.

PR #1005, opened 2026-09-12
Distinguish an unresolvable inline type name from a duplicate one

### LINKED ISSUE #997: An unparseable inline annotation is either silent (#:) or blamed on a duplicate declaration (@rbs)
## What to build

Two failure modes of the same mistake — naming a Rigor refinement where an RBS type belongs, in a `.rb` file. Neither depends on how #996 rules on an inline dialect: both are wrong answers to an unparseable annotation today, and both stay wrong whatever spelling Rigor ends up recommending.

**1. The `# @rbs` tag form degrades the whole class, behind a remediation message for a different cause.**

```rb
class ProbeZZ
  # @rbs g: finite-float
  def probe(g)
    g.to_s
  end
end
```

RBS cannot resolve `finite-float` as a type, the class's definition build fails with `RBS::NoTypeFoundError: Could not find finite`, and Rigor keeps the class KNOWN — so every call into it, real methods and typos alike, reads `Dynamic[top]`. The `rbs.coverage.definition-build-failed` warning does fire and does say the run is quieter than it should be, which is the right instinct. Its remediation is wrong for this cause: it says "Two signature sources declare the same member; remove the duplicate declaration (`rbs validate`)", which is the duplicate-declaration case. Here there is one declaration and it names a type that does not exist. A reader following the message looks for a duplicate that is not there.

The diagnostic should distinguish the failure kinds it is reporting — at minimum, an unresolvable type name versus a colliding declaration — and name the offending token (`finite`) with its position, which the underlying `RBS::NoTypeFoundError` already carries. If the token is a known Rigor refinement name, saying so is what turns the message into a fix.

**2. The `#:` annotation form is dropped in silence.**

```rb
class BadRefProbe
  #: (finite-float) -> String
  def show(f)
    f.to_s
  end
end
```

Measured at `568138c2` with `rigor annotate`: `def show(f)  #=> Dynamic[top]`. The signature never applied, the body lost its parameter type, and **no diagnostic of any kind fired** — `rigor check` on this file reports only the unrelated `rbs.coverage.missing-gem` info. Compare the tag form above, which at least warns.

Under ADR-93 an inline annotation is a live contract the engine believes, so an annotation the engine could not parse must not be indistinguishable from one that was never written. A typo in a `#:` line is the same silent failure: the author thinks they declared a signature, and nothing says otherwise.

## Acceptance criteria

- [ ] A `#:` / `# @rbs` annotation Rigor cannot parse or ingest produces a diagnostic naming the file, the line, and what could not be read — no silent drop
- [ ] `rbs.coverage.definition-build-failed` distinguishes an unresolvable type name from a colliding declaration, and its remediation matches the cause it reports
- [ ] The unresolvable-name message names the token and its position; when the token matches a Rigor refinement name, it says where that name IS valid (`%a{rigor:v1:…}` today, per `docs/manual/16-rbs-extended-annotations.md`)
- [ ] The duplicate-declaration case keeps its current message and stays regression-tested — this is a split, not a replacement
- [ ] Both repro files above are fixtures, with the diagnostic identifiers and message shapes pinned
- [ ] `make verify` green; `make check` / `make check-plugins` clean; `docs/manual/04-diagnostics.md` updated if a message or id changes

## Blocked by

None - can start immediately. Split out of #996, which rules on the inline spelling itself; this issue is independent of that ruling.

zonuexe:
## Summary

Two failure modes of the same mistake — naming a Rigor refinement where an RBS type belongs, in a `.rb` file.

**The `# @rbs name: TYPE` tag form** let the class's definition build fail with `RBS::NoTypeFoundError`, reported as `rbs.coverage.definition-build-failed`, but the closing advice always read "Two signature sources declare the same member; remove the duplicate declaration" — the wording for a genuine collision. There is one declaration here, naming a type that does not exist. `DiagnosticAggregator#definition_build_advice` now branches on the failure kind: `NoTypeFoundError` names the unresolved token and the position the error itself carries (`RbsLoader#unresolved_type_name_detail`), and when the token is the truncated head of a registered `Builtins::ImportedRefinements` name — upstream's own type parser drops everything after the first hyphen before the token ever reaches RBS — it names the `%a{rigor:v1:...}` spelling that is valid today. Every other failure kind (the two duplicated-definition errors, `GenericParameterMismatchError`, `RecursiveAncestorError`, …) keeps the original wording, which is accurate for it.

**The `#: (...) -> T` annotation form** was worse: upstream's tolerant parser never raises on an unparseable type, it records a `SyntaxErrorAssertion` tree and moves on, so the signature vanished with no diagnostic anywhere — `rigor check` reported nothing beyond an unrelated `rbs.coverage.missing-gem` info. `rigor-rbs-inline`'s existing WD12 "parsed but not honoured" channel (the one that already reports the `module-self` spelling mismatch) now also recognises a `SyntaxErrorAssertion` and reports it there, naming the real `.rb` line and the text that failed to parse.

Both fixes are additive: the duplicate-declaration message keeps its current wording and stays regression-tested, and the new detail rides the SAME per-class tuple `rbs.coverage.definition-build-failed` already caches, so it survives the ADR-54 environment cache without depending on any `RBS::Location` that cache normally degrades (verified cold vs. warm below).

## Before / after (measured at 568138c2 → this branch)

`# @rbs g: finite-float` above a `def`:

Before: `... First failure: RBS::NoTypeFoundError on \`finite\`. ... Two signature sources declare the same member; remove the duplicate declaration (\`rbs validate\`) to restore type coverage.`

After: `... First failure: RBS::NoTypeFoundError on \`finite\`. ... \`finite\` at \`virtual:rbs-inline:...:4:14...4:20\` does not name a type RBS can resolve; it is the start of the Rigor refinement \`finite-float\`, which is not valid in an RBS type position. Write it as an \`%a{rigor:v1:param: name is finite-float}\` (or \`return:\` / \`assert\`) annotation alongside the plain RBS type instead — see docs/manual/16-rbs-extended-annotations.md — to restore type coverage.`

`#: (finite-float) -> String` above a `def`:

Before: nothing beyond an unrelated `rbs.coverage.missing-gem` info; `rigor annotate` shows `Dynamic[top]`.

After: `plugin \`rbs-inline\` parsed an inline-RBS annotation in this file but did not honour it: a \`#:\` annotation on line 2 did not parse as an RBS method type or type (\`(finite-float) -> String\`) and was DROPPED — the method types as if the annotation had never been written, not merely as if its type were wrong. Fix the RBS syntax to restore it. The file's other annotations are unaffected. [source-rbs-annotation-not-honoured]`

## Test plan

- [x] `make lint` — clean
- [x] `bundle exec rspec spec/rigor/environment/rbs_loader_spec.rb spec/rigor/analysis/runner/diagnostic_aggregator_spec.rb spec/integration/plugins/rbs_inline_plugin_spec.rb spec/integration/definition_build_failed_spec.rb spec/rigor/plugin/source_rbs_synthesis_reporter_spec.rb spec/rigor/analysis/incremental_session_spec.rb spec/rigor/analysis/runner/pool_coordinator_spec.rb spec/rigor/bleeding_edge_spec.rb` — 454 examples, 0 failures
- [x] `make check` / `make check-plugins` — clean
- [x] `make docs-check` — clean
- [x] Manually reproduced both issue fixtures cold and warm (env cache cleared / reused) and confirmed identical output either way

Closes #997


zonuexe:
Orchestrator review: **`make verify` green** (exit 0), `git diff --check` clean. Verified both failure modes myself against the repro files, plus a well-formed `%a{}` control that stays silent.

The tag form now reads `` `finite` at `…:7:34...7:40` does not name a type RBS can resolve; it is the start of the Rigor refinement `finite-float`, which is not valid in an RBS type position `` with the `%a{rigor:v1:param: name is finite-float}` spelling — the misattributed duplicate-declaration remediation is gone. The `#:` form now reports `source-rbs-annotation-not-honoured` naming line 2 and the exact text that failed to parse, where it was previously silent.

Two observations, neither blocking:

- The `#:` notice is `:info` and anchored at `file:1:1` while its text names line 2. That follows the existing ADR-32 WD12 channel, which is a defensible call — but a dropped signature is a stronger event than most `:info` rows, and a reader scanning positions will land on line 1. Worth a second opinion before this is the shipped shape.
- The tag-form message still opens with "Conflicting signature file(s): …" before the new diagnosis. For this cause there is exactly one signature file, so the lead-in is still slightly off even though the remediation that follows is now correct.

Staying Draft.

PR #1006, opened 2026-09-12
Absorb a union arm that another arm contains element-wise

### LINKED ISSUE #994: Same-arity tuple union arms are not absorbed: [Float, String] | [Float[0.0..], String] survives the join
## What to build

A union of same-arity tuples does not absorb an arm that another arm already contains. An `if` / `elsif` / `else` whose three arms return the same tuple shape, differing only by a `FloatRange` narrowing on one element, joins to a three-arm union where one arm is the true answer.

`docs/type-specification/imported-built-in-types.md` already states the absorption for direct union members: "In a union, a `Float` member absorbs any `FloatRange` member and a `FloatRange` member absorbs one it contains, so the join after a guard names one set." That rule fires for a scalar join and does not reach inside a `Type::Tuple`, so the same two types wrapped in a tuple of equal arity survive side by side.

What this slice needs first is the **ruling**, because the neighbouring normalization rule points the other way: `docs/type-specification/normalization.md` deliberately keeps `1 | Integer` un-collapsed, since a value-pinned member records reachable-value provenance that collapsing would erase. The question to settle is whether element-wise containment collapses tuple arms of equal arity, and whether a value-pinned element (`[1, String] | [Integer, String]`) is excluded from that collapse the same way the scalar rule excludes it. Decide it, record it in `normalization.md`, and implement what was decided.

The cost of leaving it is not a false positive — the union is extensionally correct — but it is paid twice: the displayed type in `rigor annotate`, the Playground's inline types, and `rigor type-of` all show three arms where one is the answer, and every downstream consumer that walks union members does so three times per join.

## Repro

```ruby
class Foo
  # @rbs num: Float
  def f(num)
    if (num > 0)
      [num, num.to_s]
    elsif (num < 0)
      [num, num.to_s]
    else
      [num, num.to_s]
    end
  end
end
```

`rigor annotate` at `568138c2`:

```
    end                #=> [Float, String] | [Float[...0.0], String] | [Float[0.0..], String]
  end                  #=> :f
```

Arms two and three are both contained in arm one. The scalar control on the same run collapses correctly: the same method returning `num` rather than `[num, num.to_s]` types as `Float`, not `Float | Float[...0.0] | Float[0.0..]`. The method-level return is already widened to `[Float, String]`, so this is the in-body join, not the summary.

Also visible in the Playground on this snippet (`https://rigor.typedduck.fail/playground/`), which is where it was reported from.

## Acceptance criteria

- [ ] `normalization.md` states whether element-wise containment collapses same-arity tuple arms, and what happens when the contained element is value-pinned — the ruling lands whichever way it goes
- [ ] If the ruling is to collapse: the repro's `end` types as `[Float, String]`, with specs for the `FloatRange`, `IntegerRange`, and `Refined` element cases
- [ ] A spec pins the excluded case named by the ruling (value-pinned element, or differing arity), so the collapse cannot widen later without a spec change
- [ ] The self-check and survey-corpus diagnostic counts are unchanged — a collapse that is extensionally equal must not move a single diagnostic; if one moves, something else is wrong

## Blocked by

None - can start immediately.

zonuexe:
Closes #994

## The ruling

**Element-wise containment does collapse same-arity tuple arms, and it does so by lifting the union's own member-absorption relation one level down rather than by introducing a new one.** A value-pinned element is excluded, exactly as the scalar rule excludes it.

The two spec documents pull in opposite directions on purpose, and the resolution is to notice that they are the same rule at two depths. `imported-built-in-types.md` says a `Float` union member absorbs a `FloatRange` member, "so the join after a guard names one set" — that is a statement about the *display and consumer cost* of carrying several readings of one value set, and it is paid identically whether the `Float` sits in the union or inside a tuple in the union. `normalization.md` says `1 | Integer` stays un-collapsed — that is a statement about *provenance*, about a member that records a reachable exact value. Neither statement is about tuples. What decides the tuple case is which relation the element-wise clause is allowed to consult.

So the rule is: an arm is absorbed when, at every position, the element is equal to or absorbed by the other arm's element **under this same list**. The clause grants no absorption of its own; it asks the existing list again, one level down. Every named case then has a determined answer rather than an ad-hoc one:

| element | collapses? | because |
| --- | --- | --- |
| `FloatRange` contained in `Float` or in a wider range | yes | the direct union already collapses it (ADR-109 WD5) |
| `IntegerRange` | no | the direct union keeps `Integer \| Integer[0..5]` — both Integer edges narrow |
| `Refined` | no | the direct union keeps `String \| lowercase-string` |
| value-pinned (`[1, String] \| [Integer, String]`) | no | the direct union keeps `1 \| Integer`; the lift cannot outrun it |
| dynamic-origin | no | ADR-83 keeps a `Dynamic` arm distinct |
| `top` / `bot` element | yes | `top \| T = top`, `T \| bot = T` |
| element already listed in a `Union` element | yes | flatten-and-dedupe read one level down |
| nested tuple | yes, at any depth | the relation is recursive by construction |
| differing arity | never | `[A]` and `[A, B]` have disjoint inhabitants |
| `Tuple` beside `Nominal[Array]` | never | different carriers, not two readings of one spine |
| two mutually incomparable arms | both survive | an arm is dropped only when some *other* arm absorbs it |

The tempting alternative — "arm A is absorbed when arm B accepts it" — is wrong, and usefully so: `Tuple#accepts` is already element-wise, and `Integer` accepts `1`, so acceptance-based containment would collapse precisely the case `normalization.md` protects. The narrow relation is the point.

**The same reasoning reaches `HashShape`**, so it is ruled and implemented too, but only over an *identical spine*: same key set, same extra-key policy, same required / optional / read-only classification. Two shapes that disagree on any of that describe differently-shaped hashes, so their value types are never compared. It does not reach `Nominal` type arguments: `Array[T]` has no fixed spine and its element type is a declared generic parameter, not a position.

The ruling is recorded in `docs/type-specification/normalization.md` § "Member absorption", in the same commit as the behaviour, with a cross-reference from the `Float[...]` row of `imported-built-in-types.md`.

## The repro

`rigor annotate` before:

```
    end                #=> [Float, String] | [Float[...0.0], String] | [Float[0.0..], String]
```

after:

```
    end                #=> [Float, String]
```

## Nothing moved

The collapse is extensionally neutral, so no diagnostic may move. Holding the file set identical across arms (only `lib/rigor/type/combinator.rb` swapped), `rigor check --no-cache lib spec plugins examples apps` emits **63,779 diagnostics before and 63,776 after, and the only three that differ are the three `assert_type` assertions the new fixture adds** — the ones that are supposed to start holding. `make check` and `make check-plugins` output is byte-identical to the baseline.

## Cost on the join hot path

Measured, since every inference joins. Isolated `Combinator.union` over a two-member scalar union costs ~10% more — ~140ns on a ~1.35µs call, three alternating reps of 2M unions, reproducible to ±1% — for the widened carrier-kind guard that now admits `Tuple` and `HashShape` alongside `FloatRange`. Two cheaper guard spellings were measured and neither helped, so the readable one stayed. At run level it is invisible: `rigor check lib` wall time (12.70 / 12.90 / 13.19s before vs 12.72 / 13.21 / 14.67s after) and peak RSS (median 504MB vs 500MB) stay inside this machine's run-to-run noise.

## Verification

Run inside the Nix flake. `make verify` was **not** run — parallel lanes, concurrent full gates OOM this machine.

- `make lint` — clean, 1251 files, no offenses
- `make check` — clean, output byte-identical to baseline
- `make check-plugins` — clean, output byte-identical to baseline
- `make docs-check` — 464 examples, 0 failures
- `bundle exec rspec spec/integration` — 2098 examples, 0 failures (vacuous-snapshot count unchanged at 11)
- `bundle exec rspec spec/rigor/type spec/rigor/inference spec/docs` — 4390 examples, 0 failures
- Must-fail controls: on the unmodified combinator the new fixture reports exactly the 3 collapse assertions as mismatches, and `combinator_spec.rb` fails exactly the 6 must-collapse examples while all 11 exclusion examples pass in both arms — so the exclusion specs pin behaviour rather than passing vacuously.


zonuexe:
Orchestrator review: **`make verify` green** (exit 0), `git diff --check` clean.

The ruling is the right one and it is argued from the right place — the element-wise clause consults the union's own member-absorption relation one level down rather than introducing a new one, so every exclusion follows instead of being stipulated. That is what makes `normalization.md` and `imported-built-in-types.md` both true at once rather than one of them yielding.

Verified the two poles myself rather than reading the specs:

```
# @rbs f: Float — three arms, FloatRange elements
    end              #=> [Float, 2]

# @rbs n: Integer — one arm value-pinned
    end              #=> [1, "x"] | [Integer, "x"]
```

The first collapses, the second does not, which is exactly `1 | Integer` surviving one level down. The issue's original repro now types as `[Float, String]`.

The 63.8k-diagnostic sweep with only the three new fixture assertions moving is the right shape of evidence for an extensionally-equal change, and the join micro-benchmark being +10.5% on an isolated 1.35µs call while invisible at run level is worth keeping in the PR body — the next person to touch that guard will want the number.

Staying Draft.

Issue #898, opened 2026-09-09
flow.unreachable-clause fires on a class object extended with a core module: the singleton model ignores `extend`


zonuexe:
Found while widening #657's decline to the remaining positive-edge carriers. Deliberate residue of that fix, not a regression from it.

`Rigor::Inference::Narrowing.narrow_singleton_to_class` (`lib/rigor/inference/narrowing.rb`) approximates `Singleton[Foo]` as "an instance of `Class`" and asks the class ordering about `Class` alone. That model is silent about every module a class object was extended with, so a `Bot` collapse on a **module** target rests on evidence the model never had.

## Repro

```ruby
class Widget
  extend Comparable
end

k = Widget
case k
when Comparable then k.clamp(1, 2)
end
```

```
info: unreachable `when Comparable': `k' can never be Comparable here (the flow proves the subject disjoint) [flow.unreachable-clause]
```

`Widget.is_a?(Comparable)` is `true` in MRI — the `extend` puts `Comparable` in `Widget`'s singleton ancestry.

## Why #657's fix does not cover it

The decline landed there keys on an `:unknown` ordering. `Class` vs `Comparable` is not unknown: both are in RBS and neither includes the other, so the ordering answers `:disjoint` and the collapse still fires. The project-module form (`extend Meta` where `Meta` is project source) IS covered, because `Class` vs `Meta` is `:unknown` — which makes the surviving false positive the one where the environment knows more, not less.

## Shape of a fix

The evidence a `Bot` needs here is the singleton's own ancestry, and the model does not carry it. Two candidates, both worth weighing against the false-positive-first rule:

1. **Decline on every module target.** `narrow_singleton_to_class` collapses only when the target resolves as a *class* (where "a class object's class is `Class`" is real evidence and `Widget.is_a?(Integer)` is genuinely false), and joins to Dynamic for a module target. Cheap and complete; it retracts every `case SomeClass when SomeModule` unreachable report, which the model cannot establish anyway.
2. **Model the singleton ancestry.** Feed in-source `extend` (and `class << self; include`) into the ordering the singleton path asks, so `Widget` vs `Comparable` answers `:subclass`. Precise, but it derives a positive edge from project source and needs its own guard against conditional `extend`.

Fix 1 alone is enough to close the false positive.

## Gate

[...]

Issue #667, opened 2026-09-02
Cross-file constant truthiness withholding does not survive a copy: m = AppConfig::MODE then m == :production still warns


zonuexe:
Split out of [#644](https://github.com/rigortype/rigor/issues/644) during its adversarial review.

#644 publishes cross-file value constants and withholds them as fuel for `flow.always-truthy-condition` — a condition that folds only because of a project-published constant the reading file does not itself assign must not warn, because branching on a config constant is the idiom whose whole point is that it is constant.

The withholding tracks the constant through a **direct read** and through one implicit-self predicate hop. It does not track a value COPIED out of the constant first, so these still warn on correct config code:

```ruby
m = AppConfig::MODE
m == :production                      # local copy

@mode = AppConfig::MODE               # in initialize
@mode == :production                  # ivar copy

MODE2 = AppConfig::MODE               # same-file alias of a foreign value
MODE2 == :production
```

All three are silent on master (the value was `Dynamic`, so nothing folded) and warn with #644 landed. Each is an ordinary way to read configuration.

## Why it was split

Direct reads and the predicate hop are answerable from name sets (`Scope::DiscoveryIndex`), which is what #644 shipped. A copy is not: it needs the *provenance* of a value to survive an assignment, so the withholding travels with the value rather than being re-derived from the expression's shape at the point the rule looks.

## Fix direction

Stamp the provenance on the value and let it flow. [ADR-58](docs/adr/58-ivar-field-typing.md)'s `r = @x` stamping is the precedent, and the declaration-sourced mark (`docs/internal-spec/inference-engine.md` § "Carrier and discipline": "no diagnostic MUST fire *from* it") is the existing shape for "precise enough to dispatch on, not evidence enough to report from".

Two cautions, both from #644's review:

- The ADR-58 mark's `Set[[kind, name]]` is **shared with ADR-67's oppositely-joined taint**, and the spec treats a third establishing transition as the *Non-transitivity* decision. #644 deliberately avoided adding a kind to it for that reason. Read that before reusing it.
- It is flow state keyed on a rebindable binding, which is right for this (a copy IS flow) where it was wrong for #644's project-wide fact.

Scope: the alias case (`MODE2 = AppConfig::MODE` at another file's top level) may fall out of the census instead — a published constant assigned FROM a published constant could inherit the mark at publication time, with no flow machinery at all. Check that cheaper route first.

## Gate

[...]

Issue #656, opened 2026-09-02
Constant paths never resolve their first segment through ancestors: A::B inside P::Guard answers top-level A::B where Ruby answers P::Base::B


zonuexe:
Found by the adversarial review of [#635](https://github.com/rigortype/rigor/issues/635). **Pre-existing on master**, on every guard shape and on plain constant reads alike.

A constant path's **first segment** is resolved through the lexical nesting only. Ruby also resolves it — and the segments after it — through the **ancestors** of the enclosing class, so a path whose head is provided by a superclass or an included module gets the wrong class, and the wrong class is then reported against.

## Repro

Ruby ground truth: inside `P::Guard`, `A::B` resolves `A` lexically to `P::A`, then looks `B` up through `P::A`'s ancestors, reaching `P::Base::B`.

```
F2 ancestor path: A::B resolves to: P::Base::B
```

Rigor answers the top-level `A::B` instead, on both the read and every guard shape. With RBS declaring both classes, `rigor check` fires three times on correct code:

```
f2.rb:26: error: undefined method `base_b' for A::B [call.undefined-method]
f2.rb:31: error: undefined method `base_b' for A::B [call.undefined-method]
f2.rb:35: error: undefined method `base_b' for A::B [call.undefined-method]
```

(the read, `is_a?`, and `when` positions respectively). This is a **wrong name**, not lost precision — the analyzer names a real but different class and then type-checks against it.

## Mechanism

Two resolvers share the blind spot:

- `Inference::Narrowing#resolve_class_name_lexically` walks nesting prefixes and tries `<prefix>::<full path>`; it never resolves the head segment on its own and then continues from what it found.
- `Reflection.resolve_constant_type` step 3 tries `<ancestor>::<full path>` — again the whole path under an ancestor, not the head segment through the ancestor followed by the remaining segments inside the resolved constant.

Both work for a path fully owned by one prefix. Neither models Ruby's actual two-phase lookup: resolve the first segment through nesting *then ancestors*, then resolve each subsequent segment within the constant the previous segment produced (searching that constant's own ancestors in turn).

## Fix direction

Make path resolution segment-wise rather than whole-path-under-a-prefix: resolve segment 1 through the existing ladder (nesting, then ancestors, then top level), then walk the remaining segments inside the resolved owner, consulting that owner's ancestors at each step. The candidate must still be adopted only when the environment or discovery actually knows it, with the as-written spelling as the fallback, so an unresolvable path stays gradual rather than becoming a guess.

[...]

PR #1010, opened 2026-09-13
Arity-check a method the project defines in source and declares nowhere

### LINKED ISSUE #992: Arity-check source-defined methods that carry no declaration: the DiscoveryIndex records no parameter shape

zonuexe:
Closes #992

## Ruling: on by default, no configuration key

The survey corpus is 34 targets, 18 of them run on their configured paths, with subsets for `gitlab` (`app/models lib/gitlab/ci`) and `rails` (`activemodel/lib activerecord/lib`). All runs used `--no-cache`, with the baseline built by toggling the changed files in this worktree. In the final build it adds **zero** `call.wrong-arity` firings and changes no other diagnostic count on any target. An instrumented intermediate build counted about 8,400 call sites that reached a source envelope across the corpus. Only four fell outside their envelope, and each one led to a decline this PR now carries (table below). Every decline listed here has a spec whose fixture fires once that decline is knocked out, and each was checked by knocking it out. I found no evidence that justifies a gate, and a gate would hide the rule from the day-one projects the issue is about.

## One source of truth for an undeclared `def`'s envelope

`Scope::DiscoveryIndex#discovered_parameter_envelopes`: `{class => {[kind, name] => [min, max, required_keywords] | :opaque}}`. It is written by `ScopeIndexer#record_method`, the same recorder that now writes `discovered_methods`. That means every name the existence table learns from `attr_*`, `alias`, `define_method` or a `Struct`/`Data` member lands here as `:opaque`, and only a `def` records a real envelope (`Source::ParameterEnvelope.of`). Every fold uses `ParameterEnvelope.merge`: equal envelopes stay, anything else becomes opaque, and opaque absorbs. That applies to reopenings in one file, the cross-file pass, the ADR-85 bundle fold, compact-header re-keying and the per-file join with the project seed. No path keeps one of several shapes.

Both undeclared lanes read this table:
- **No member** (RBS-unknown class, or a class whose RBS is the project's own `sig/` but lacks the method) goes to the table.
- **`%a{rigor:v1:inferred-signature}` member** (the `rigor-rbs-inline` skeleton) also goes to the table. The synthesized member's arity is never read. The two #999 regression arms (`Bar#g`, `Helper#helper`) now fire with exactly the messages the no-member lane produces.

A bundled (core/stdlib/gem) RBS declaration without the member is an ADR-17 monkey-patch and declines. A declared member keeps the existing RBS path unchanged.

## Declines (each has its own spec; each spec fails when its decline is knocked out)

Stage 1 is the walk to the nearest level that records the name. The class and its project mixins must agree there, because `include` and `prepend` share one table.
- The receiver is a project module, or `self.class` of a project-module value.
- It is an ADR-26 open receiver, or a plugin-typed call site.
- Required keywords on the def. Keyword arity is out of scope, mirroring `compute_arity_envelope`.
- No whole-project discovery pass seeded the scope (the LSP `prebuilt:` runner).
- The nearest level records the name more than once with different shapes: a reopening in the same or another file, a conditional pair of defs, an included or prepended project module.
- The name also comes from `define_method` / `define_singleton_method`, `alias` (either name), `alias_method`, `undef`, or any class-body macro naming it (`memoize :f`, `memoize def f`, `def_delegators`). Visibility macros are the only exception.

Stage 2 runs only when the count falls outside the envelope, since everything here can only withhold.
- `method_missing` / `respond_to_missing?` anywhere in the chain.
- Dynamic surface: `class_eval`-family calls, a computed `define_method`, `send`, a non-constant mixin, a project mixin helper (`prepend_mod_with`), `Const.class_eval` / `Const.include` from outside, `refine Const`. A `class_eval` / `define_method`-family call inside any method body (`def self.inherited(sub) = sub.class_eval …`) counts too.
- A module that some method body passes to `extend` and that defines the name.
- A mixin the project does not declare that is not an RBS-known module lacking the name.
- A `pre_eval:` patch or a plugin synthetic method.
- The owner makes the method non-public.
- A project subclass records a different or opaque definition, or is dynamic.
- The receiver's class shares its last name segment with another known class the call site's nesting can reach (load-order dependent).

Existing declines stay ahead of all of this: the call-site shape (splat, `**`, keywords, `...`), `unenumerable_receiver_shape?` (#879), and the ADR-67 WD6b inferred-parameter receiver. Each has a spec here. A block-pass is counted by its positionals, because `f(&blk)` against `def f(a)` raises too.

Pinned as **in scope**: `super` / `yield` in an override, an unknown gem superclass under the class's own def, a `SimpleDelegator` subclass's own def, `include Comparable`.
Pinned as **out of scope**: constructors through `Class#new` (`initialize` is never read as `.new`).

## Corpus adjudication

| Target | Before | After | Call | Verdict |
| --- | --- | --- | --- | --- |
| gitlab | 0 | 0 | `IntegrationsHelper.integration_event_description(event)` (`app/models/concerns/integrations/base/integration.rb:206`) against `def integration_event_description(integration, event)` + `extend self` | **Real bug.** Dead code: `event_description` has no callers. Declined anyway, because the helper does `prepend_mod_with('IntegrationsHelper')` from an `ee/` tree the analysed paths do not reach. |
| tdiary-core | 0 | 0 | `PStore::new(filename).transaction do` inside `TDiary::IO::PStore#transaction(date)` (`lib/tdiary/io/pstore.rb:23`) | **Real bug**: `PStore` resolves to the class itself. Declined by the load-order check, since `::PStore` is also known. |
| tdiary-core | 0 | 0 | `PStore.new(...).transaction do` in `TDiary::IO::PluginPStore` (`lib/tdiary/io/plugin_pstore.rb:21`) | **Would have been a false positive**: under the default `io_class`, `TDiary::IO::PStore` is never loaded and `PStore` is the stdlib class. The load-order check was added for this. |
| concurrent-ruby | 0 | 0 | `Channel.select do \|s\|` in `Promises::FactoryMethods::OldChannelIntegration` against `Concurrent::Promises::Channel.select(channels, timeout = nil)` | Load-order dependent: `Concurrent::Channel.select(*args)` is what the author meant (`# noinspection RubyArgCount`). Declined by the same check. |

Every other target: before 0 → after 0 for new firings. mastodon (2), net-ssh (2) and protobuf (1) keep their existing RBS-path firings, unchanged.

## Caches

- **Run cache (ADR-45)**: diagnostics only. The engine-source slot already moves on this lib edit. Spec: warm == cold through a real `Cache::Store`, plus a warm run after another file adds `memoize :f`.
- **ADR-85 seed bundle**: carries `parameter_envelopes`. `IncrementalSnapshot::SCHEMA` 22 → 23. Spec: a bundle-served fold equals the cold walk, round-tripped through `Marshal`.
- **ADR-89 WD1 declaration signature**: includes the joined table. Spec: adding only `memoize :f` moves it.
- **ADR-46 edges**: stage 1's reads are withheld. A fitting call records only the owner defs' symbol edges plus name-keyed negatives for the levels it passed, so `dependency_recorder_spec`'s "method-call deps, not ancestry" contract holds. Firing and opaque verdicts replay the walk. Specs: incremental rechecks for narrowing the owner, a nearer level gaining a def, and a wrap in another file on both a plain class and a `sig/`-declared one.
- Also plumbed: `Runner` seed tables, `ProjectPrePasses::Discovery`, and `Protection::DiscoverySeed`. `Cache::Descriptor::SCHEMA_VERSION` did not move, because no persisted descriptor shape changed.
- **Known residuals**: `--incremental` does not see an ADDED subclass file that overrides the method until a full run. This matches `method_defined_on_known_subclass?`. The discovery walk sees a `class_eval` or `extend` on a non-constant receiver only as a mark on the lexical class or the extended module. Reopenings outside the analysed paths are invisible, as they are for every rule.

## Docs

- `diagnostic-policy.md`: the `call.wrong-arity` row now requires the source witness and lists its declines. This reverses the "a project `def` alone MUST NOT witness" sentence.
- `internal-spec/inference-engine.md`: the new table's contract.
- `internal-spec/cache.md`: Payload SCHEMA 23, and the declaration signature.

`sig/rigor/scope.rbs`: `rigor sig-gen --print` / `--diff` skips both new readers (untyped return), so their rows are hand-written in the file's existing `sig-gen gap:` form.

## Verification

The commands and their results are in the session report. `make verify` ran once, after the changelog fragment commit.


zonuexe:
Orchestrator review of `1896441a`: **CI green**, and the agent’s `make verify` exit 0 stands. I re-checked the behaviour myself with `--no-cache` in a scratch project rather than from the specs.

**Fires, correctly, across files:** `foo.g` against `def g(a, b = 1)` (given 0, expected 1..2), `foo.g(1, 2, 3)` (given 3), `foo.h` against `def h(a, *rest)` (given 0, expected 1..Infinity), and `Helpers.help` with no argument against a `module_function` `def help(a)` — that last one raises `ArgumentError` at runtime, so it is a true positive, not a decline that leaked.

**Silent, correctly:** a `define_method` target, a class with `method_missing` / `respond_to_missing?`, a name reopened with two different shapes, an `alias`ed name, and — the one that matters most in practice — `b.x` where `b` is declared `Base` and a `Child < Base` overrides `x` with a different arity.

One conservative miss, acceptable and worth knowing: `Base.new.x` with the wrong arity also stays silent, because the subclass decline applies to `Nominal[Base]` even when the receiver is a literal `Base.new`. That is a false negative, not a false positive.

`expected 1..Infinity` is not new: the RBS lane already renders `Kernel.format` and `Array#insert` that way.

The three hand-written `sig/rigor/scope.rbs` rows use `Hash[untyped, untyped]` and `sig-gen gap: #992` markers. That follows the convention the existing #656 / #667 / #898 markers set, so nothing needs to change here — but that convention disagrees with what the ADR-107 G3 gate says a marker must point at. Filed separately as #1011.

Staying Draft until the user decides on landing it default-on.

PR #1018, opened 2026-09-14
Read the same-line %a{} inline annotation forms

### LINKED ISSUE #998: Rigor silently drops the same-line %a{} annotation forms that rbs built-in reader accepts
## What to build

Rigor's inline reader silently drops the two `%a{}` spellings that rbs's own built-in reader accepts, and it drops them the way ADR-32 WD12 forbids: without a diagnostic, and in one case while keeping a signature it has quietly emptied.

Measured at `568138c2` with rbs 4.2.0 / rbs-inline 0.14.0, both readers run over the same source (probe script and the full twelve-row table: `docs/notes/20260912-inline-refinement-carrier-probe.md`):

| Spelling | `rbs-inline` gem — Rigor's reader | rbs `RBS::InlineParser` |
| --- | --- | --- |
| `# @rbs %a{rigor:v1:return: non-empty-string}` on its own line, then `# @rbs return: String` | annotation lands on `def name: () -> String` | `AnnotationSyntaxError: expected a token pARROW` — annotation lost, plain type kept |
| `# @rbs %a{rigor:v1:return: non-empty-string} () -> String` on one line | annotation lands, **method type dropped** → `def name: () -> untyped`, no diagnostic | `() -> String` with the annotation attached, no diagnostic |
| `#: %a{rigor:v1:return: non-empty-string} () -> String` | **whole line dropped** → `def name: () -> untyped`, no diagnostic | `() -> String` with the annotation attached, no diagnostic |

`%a{pure}` — rbs core's own annotation — splits identically, so this is not about the `rigor:v1:` payload.

Two things are wrong on Rigor's side, independent of which spelling the manual ends up recommending:

1. **The drop is silent.** ADR-32 WD12 says a parsed-but-unhonoured annotation is never swallowed, and `docs/type-specification/overview.md` claims 100% compatibility with rbs-inline syntax. A line that is valid rbs inline syntax, that the ecosystem's long-run reader (ADR-94) accepts, must not vanish without a word — and the second row is worse than silence, because it replaces a declared `() -> String` with `() -> untyped` and says nothing.
2. **The same-line forms should reach the environment with their annotation attached.** They are the forms `RBS::InlineParser` accepts, so they are the forms that survive Rigor's eventual migration to it. The injection point is the plugin's synthesis seam — `plugins/rigor-rbs-inline/lib/rigor/plugin/rbs_inline.rb`, where the existing `default_type` marker and the `#:nodoc:` rewrite already sit.

Once both forms reach the environment, `docs/manual/16-rbs-extended-annotations.md` documents both and recommends the same-line one.

## Context

Surfaced by the ADR-111 probe (`docs/adr/111-inline-refinement-carrier.md`, WD5), which was measuring something else: whether a Rigor refinement in a `.rb` file needs a carrier of its own (#996). The answer turned on what each reader sees, and the measurement found this. The manual's compatibility claim — "any other RBS tool preserves or ignores the annotation" — is not true of the built-in reader as measured, and ADR-111's recommendation is stated against that gap rather than around it.

This is the same class of defect as #997 (an annotation Rigor cannot read disappearing in silence), reached from the other side: here the annotation is well-formed and Rigor's reader is the one that loses it.

## Acceptance criteria

- [ ] `# @rbs %a{…} () -> T` and `#: %a{…} () -> T` reach the environment with both the method type and the annotation attached
- [ ] No inline construct is dropped without a diagnostic — the ADR-32 WD12 rule holds for every row of the probe table
- [ ] The own-line form keeps working unchanged; this adds spellings, it does not move the documented one
- [ ] The probe note's table is re-run and updated in the same PR, so the record shows the new Rigor column
- [ ] ADR-32 WD11's lost-construct list gains the row, and `docs/manual/16-rbs-extended-annotations.md` documents both forms with the same-line one recommended

## Blocked by

None - can start immediately. Independent of #996's ruling and of #997.

zonuexe:
Closes #998.

`# @rbs %a{…} () -> T` and `#: %a{…} () -> T` are the inline spellings rbs's built-in `RBS::InlineParser` and Steep's inline mode accept. The rbs-inline gem Rigor reads with kept the annotation and dropped the method type of the first (silently, and marked the method inferred), and turned the second into a `SyntaxErrorAssertion` — which, since #1005, drew a `source-rbs-annotation-not-honoured` info telling the author their valid line "did not parse … and was DROPPED". Measured on `19d7105d` for both `%a{rigor:v1:return: non-empty-string}` and `%a{pure}`.

## Mechanism

A post-parse repair of the gem's annotation list, in `plugins/rigor-rbs-inline/lib/rigor/plugin/rbs_inline/same_line_annotations.rb`, applied before the writer runs and to the list the WD12 notices are computed from. Not a source rewrite: the gem has no one-line spelling for annotation + method type, and splitting the comment onto two lines would shift every later line (the `def`'s comment association, the notice's line number).

Why it is exact: the gem's own `Tokenizer` delimits the leading `%a{…}` tokens; each half is re-read by the gem's own `AnnotationParser` as the own-line comment (`# @rbs %a{…}`, then `# @rbs () -> T` / `#: () -> T`) at the original line and column; the split happens only when the remainder re-reads as a well-formed method type. Otherwise the gem's original annotation is left untouched. The gem is not modified or patched; `rbs-inline --output` still drops these forms.

## Behaviour

| spelling | before (`19d7105d`) | after |
| --- | --- | --- |
| `# @rbs %a{rigor:v1:return: non-empty-string} () -> String` | `() -> untyped` + inferred marks, silent | annotation + `() -> String`, no marks; call site `non-empty-string` |
| `#: %a{rigor:v1:return: non-empty-string} () -> String` | `() -> untyped`, false "did not parse" info | annotation + `() -> String`, no diagnostic; call site `non-empty-string` |
| `#: %a{pure} () -> String` | as above | `%a{pure}` + `() -> String` |
| `#: %a{pure} (finite-float) -> String` | #997 info | #997 info (unchanged) |
| `# @rbs %a{pure} (finite-float) -> String` | silent | new info naming the discarded text |
| own-line `# @rbs %a{…}` + `# @rbs return: String` | works | unchanged |

The probe note's table gains a "Rigor after #998" column (Steep not re-run: `tool/steep/` is not installed in this worktree). Every other row is unchanged. Rows B (`# @rbs-ext …`) and H (`# @rbs n: Integer[1..10]`) are still silent in Rigor; they are outside this split, so the issue's "no row is dropped without a diagnostic" criterion is not met for them by this PR.

## Docs

ADR-32 WD11 gains the `%a{}` line-shape row and WD12 a paragraph; manual 16 documents all three spellings with the reader-compatibility table and no recommendation (ADR-111 is Proposed); the rigor-rbs-inline plugin manual gains a "Same-line annotations" section and the fourth notice cause. ADR-111 gets factual corrections only; status and recommendation untouched.

## Verification

New specs fail on `19d7105d`'s `rbs_inline.rb` (9 failures, including the must-fail malformed-`@rbs` notice), pass here. `make lint`, `make check`, `make check-plugins`, `make docs-check` clean. `make verify` not run in this lane.


zonuexe:
Orchestrator review. Checked cold (`--no-cache`) in a scratch project, not from the specs:

- **The false notice #1005 introduced is real on master and gone here.** On `19d7105d` a valid `#: %a{rigor:v1:return: non-empty-string} () -> String` reports "did not parse … and was DROPPED"; on this branch it reports nothing.
- **Both same-line forms now bind.** The call sites of `# @rbs %a{…} () -> String` and `#: %a{…} () -> String` type as `non-empty-string` here and as `"x"` (the method read as `() -> untyped`) on master.
- **A malformed same-line form is not a regression.** `#: %a{pure} () -> Strin(` now reports `rbs.coverage.definition-build-failed` naming `Strin`, where master reported "did not parse". That is exactly what master already does for the plain `#: () -> Strin(` (the gem tolerates the trailing `(`), so the annotated form now matches the unannotated one rather than getting its own treatment.

On scope: #998's "no row dropped without a diagnostic" criterion is not fully met — rows B (`# @rbs-ext`) and H (`# @rbs n: Integer[1..10]`) are still silent, as the PR body says. Those are split out to #1019, so `Closes #998` stays accurate for what #998 was filed to fix.

Issue #1003, opened 2026-09-12
A predicate guard narrows in if/else but not in the ternary spelling of the same guard


zonuexe:
## What to build

A predicate guard narrows the subject in an `if` / `else` body and does not narrow it in the ternary spelling of the same guard. The two forms parse to the same Prism node kind, so a reader has no way to see why one of them is less precise.

```ruby
class Probe
  # @rbs f: Float
  def probe(f)
    if f.finite?
      g = f          # finite-float
    else
      g = 0.0
    end

    h = f.finite? ? f : 0.0   # 0.0 | Float — the true arm did not narrow
  end
end
```

Measured at `568138c2`. The `if` form reaches `finite-float`, which is the ADR-109 WD5 truthy-edge rule working. The ternary's true arm keeps the entry type.

The join is not the explanation. `f.finite? ? f.to_s : "x"` types as `"x" | non-empty-string`, and the `if` form of the same expression gives `numeric-string` on the truthy edge — so the receiver itself is unnarrowed inside the ternary arm, before anything is joined.

`f.nan?` behaves the same way, so this is about where the narrowing facts are applied rather than about one predicate.

## Why it matters beyond precision

Every refinement that depends on a guard is reachable in one spelling and not the other, and the ternary is the spelling people reach for when the guarded expression is short — which is most guards. It also silently weakens anything built on those facts: a rule that fires only when a value is proven finite, non-empty, or non-nil simply will not see the proof through a ternary.

The fix is presumably narrow — apply the same narrowing facts when the `if` node is in its ternary form — but the acceptance criteria should be read as "the two spellings agree", not "this one predicate works", because the gap is structural.

## Acceptance criteria

- [ ] `x.finite? ? A : B` narrows the subject in `A` exactly as `if x.finite?` narrows it in the body, and the falsy arm keeps the entry type (`!(x > c)` admits `NaN`, per `docs/type-specification/imported-built-in-types.md`)
- [ ] The same holds for the predicate catalogue generally (`nan?`, `nil?`, `is_a?`, `empty?`, `respond_to?` — whatever `docs/type-specification/control-flow-analysis.md` lists), with a spec per family rather than one for the case in the repro
- [ ] `unless` in its modifier form, and `a ? b : c` nested in another condition, are checked too — if any spelling still disagrees, the issue says which and why
- [ ] A precision snapshot fixture covers both spellings side by side, so a future regression shows up as a diff rather than as a quietly wider type
[...]

Issue #521, opened 2026-08-31
OverloadSelector pins a single overload on a Dynamic discriminating argument: [true] * n types String


zonuexe:
Found by the 2026-09-01 corpus opacity sweep on numo-narray (`tool/opacity-sweep-20260901/reports/numo-narray.md`, branch `opacity-sweep-20260901`). With `ndim` typed `Dynamic`, `idx = [true] * ndim` types `idx` as **String** — `Array#*` has `(::String) -> ::String` and `(::int) -> ::Array[Elem]` overloads, and overload selection with a `Dynamic` argument pins the String-join arm instead of declining or joining. 10 direct sites there, and the wrong precise type is FP-relevant: downstream `idx[i] = j` then type-checks against `String#[]=`.

This is the same family as the overload value-pinning hijack fixed in OverloadSelector pass 2 (see `docs/notes/` prior art), one level up: a `Dynamic` argument matches EVERY overload and must not select a specific one. Expected behaviour: when the discriminating parameter position receives `Dynamic`, either answer the union of the candidate overloads' returns or fall to `Dynamic` — never a single arm's precise return. False positives outrank worst-case static reading, so this ranks above any precision lever.

**Repro sketch:** `n = [1,2][rand(2)]; x = [true] * n` — inspect with `rigor type-of`; on master x reads String.

**Gate:** `make verify` + self-check gates; corpus FP diff. A spec pinning the union/Dynamic answer for a Dynamic discriminator, plus a must-still-succeed control for literal discriminators (`[true] * 3` stays `Array[true]`-shaped) per the formatter-splat lesson (`project_rigor_formatter_strips_splat`).

Issue #313, opened 2026-08-08
Optimistic nil-free HashShape read launders through .nil? into the &&/|| polarity gate


zonuexe:
## Symptom

A closed literal hash read with a DYNAMIC key folds to the nil-free value union of its values (`0 | 1 | 2` for a rank table) — which `docs/internal-spec/inference-engine.md`'s ADR-78 WD1 paragraph declares OPTIMISTIC rather than proof, and explicitly forbids `flow.always-truthy-condition` and the `&&`/`||` `constant_value_polarity` gate from concluding on. The exclusion holds on the single-operand path but NOT through composition: the `.nil?` predicate folds the optimistic union into a bare `Constant[false]` carrying no `OptimisticOrigin` mark, and the polarity gate then reads that unmarked Constant.

Minimal reproduction (verified on master 9df52a73):

```ruby
MAP = { public: 2, protected: 1, private: 0 }.freeze

def single(a)
  x = MAP[a]
  return false if x.nil?   # silent — correct
  x
end

def double_or(a, b)
  x = MAP[a]
  y = MAP[b]
  return false if x.nil? || y.nil?   # fires "condition is always falsey" — WRONG
  x < y
end
```

`MAP[:unknown]` really is nil at runtime, so the guard is live defensive code and the diagnostic is a false positive on correct input — the exact damage class the repo's FP-first axis weighs heaviest. The `&&` spelling (`return false unless !x.nil? && !y.nil?`) misfires symmetrically as always-truthy.

## Where it bit

`check_rules.rb`'s own `visibility_reduced?` guard (`return false if parent_rank.nil? || override_rank.nil?`) — hidden until #306 lifted that file's accidental self-suppression, and now carried by a justified `# rigor:disable flow.always-truthy-condition` directive whose comment names this issue's mechanism. Removing that directive is the fix's acceptance test.

## Fix direction

Propagate the optimism mark through the predicate fold: when `.nil?` (and the sibling unary predicates) folds a value whose nil-freeness is optimistic (the #286 `OptimisticOrigin` machinery is the existing carrier for exactly this distinction), the resulting `Constant[false]` must carry the mark, and `constant_value_polarity` must decline marked constants. Note the relationship to #152 (closed): the spec warned that any widening of the polarity gate MUST preserve the optimistic-value exclusion, and the composed-operand path appears to have lost it — check the #152 landing against the spec clause while fixing.

## Provenance

[...]

PR #1020, opened 2026-09-14
Narrow a value-position and/or through the statement evaluator

### LINKED ISSUE #1016: A bare && / || used as a value does not narrow its right operand, unlike the same expression as a statement
## What happens

A bare `&&` / `||` used as a **value** does not narrow its right operand, while the same expression evaluated as a **statement** does. Measured on master `2b26e4f9`:

```ruby
class AndOrProbe
  # @rbs x: Float
  def probe(x)
    a = x.finite? && x      # value typed Float | false — want false | finite-float
    b = [x.finite? && x]    # [Float | false]
  end
end
```

The statement evaluator's `eval_and_or` gives `false | finite-float` for the same expression. The value-position typer in `ExpressionTyper` gives `Float | false`.

## Why this is one bug class, not a one-off

#1003 (fixed by #1015) was exactly this shape for conditionals: `StatementEvaluator#eval_if` narrowed each arm, while `ExpressionTyper#type_of_if` was a second typer that typed both arms in the entry scope. #1015 deleted the parallel typer and routed value-position `IfNode` / `UnlessNode` through `scope.evaluate`, so the two spellings cannot drift again. `&&` / `||` still have both implementations.

The agent that fixed #1003 did not apply the same delegation here, for a stated reason worth honouring: the and/or value typer carries the issue #313 short-circuit polarity gate, and `eval_and_or` does not. Routing the value through the statement evaluator as-is would drop that gate. So the fix is to reconcile the two — give the statement path the #313 gate, or move the gate to where both paths share it — and then have one implementation, not to paper over the value typer.

## Acceptance criteria

- [ ] `x.finite? && x` and `x.nil? || x` type identically as a statement and as a value (a call argument, an array element, a receiver, the value `rigor type-of` reports on a write)
- [ ] One implementation of and/or narrowing remains; the #313 short-circuit gate still holds on every path, with its existing specs green
- [ ] Paired specs per predicate family, statement vs value, in the style of `spec/rigor/inference/conditional_value_position_narrowing_spec.rb` from #1015 — each against a pinned expected type, so two equally wide results cannot pass
- [ ] A survey-corpus before/after with every moved diagnostic adjudicated (the #1015 PR body shows the method: toggle the changed `lib/` files in-tree, `--no-cache`, check each run's exit status and non-empty JSON)
- [ ] `make verify` green

## Blocked by

None - can start immediately. Follow-up to #1003 / #1015.

zonuexe:
Closes #1016

## Mechanism

A bare `&&` / `||` had two typers. `StatementEvaluator#eval_and_or` evaluated the right operand under the left operand's narrowing edge. `ExpressionTyper#type_of_and_or`, which answers every value position (an argument, a receiver, a literal element, the value `rigor type-of` reports for a write), typed both operands in the entry scope, so `x.finite? && x` read `Float | false` as a value and `false | finite-float` as a statement. This is the #1003 / #1015 bug class, for and/or.

Read end to end, the two typers differed in four ways, not only in the #313 gate:

1. **RHS narrowing.** The value typer typed the RHS in the entry scope.
2. **Left-operand scope threading.** A binding made in the left operand (`(y = s) && y.upcase`) was invisible to the value-position RHS (`Dynamic[top]?` against `String?`).
3. **Constant short-circuit.** Only the value typer dropped a dead RHS: `1 || s` was `1` as a value and `1 | String | nil` as a statement; `nil && 5` was `nil` against `5?`. The #313 optimistic-carrier decline guards exactly this short-circuit. The statement side needed no gate only because it never concluded.
4. **Where the #313 mark was resolved.** The value typer resolved it against the entry scope, so `(w = uniform[key]) || 5` still discarded the fallback in value position (`1` against the statement's `1 | 5`).

## Fix

The constant short-circuit and its #313 decline move into `StatementEvaluator` (`right_operand_dead?`). The decline is resolved against the left operand's post-scope, as `branch_certainty` already does for `if`. `ExpressionTyper#type_of_and_or` then delegates to `scope.evaluate` and reads only the value, like `type_of_conditional`. `left_operand_polarity`, `short_circuit_for`, `constant_value_polarity` and the typer's `optimistic_origin_for` are deleted. The dead RHS is still evaluated and still joins the post-scope, so only its value is dropped and nil-injection of a write inside it is unchanged. The gate stays `Constant`-only (#152).

Normative text: `docs/internal-spec/inference-engine.md` (the statement rule gains the short-circuit and its decline; a value-position delegation bullet for `AndNode` / `OrNode`) and `docs/type-specification/control-flow-analysis.md`.

## Specs

- `spec/rigor/inference/and_or_value_position_narrowing_spec.rb` has 20 examples. Each expression is typed in five positions (statement, `type_of`, array element, `ScopeIndexer` receiver scope, write node) against one pinned type. Families: `finite?`, `nan?`, Float comparison (the falsey edge keeps `Float`), projection through the narrowed receiver, `nil?` / `!`, truthiness, `is_a?`, `respond_to?`, `empty?`, literal `==`, nested `&&` inside `||` both ways, left-operand binding, the constant short-circuit control, and the #313 declines (`uniform[key] || 5`, `uniform[key] || key`, `(w = uniform[key]) || 5`, `missable.nil? && 5`, `!missable.nil? || 5`, the composed `x.nil? || y.nil?` guard, and a ternary on it). On the master engine, 14 of the 20 fail.
- `spec/integration/fixtures/and_or_value_position_narrowing.rb` is a precision snapshot. It diffs on 10 items against the master engine.
- `statement_evaluator_spec.rb`: `1 || "hi"` pinned the old statement-only union. It now pins the short-circuit, including that a write in the dead RHS still nil-injects. The union example uses a non-constant left.
- `provenance_spec.rb`: `type.rbs` 212 → 211. `AnonymousClassName.match?` (`class_name.is_a?(String) && class_name.start_with?(PREFIX)`) now generates its declared `bool`.
- The #313 specs (`optimistic_origin_spec.rb`, the runner's always-truthy "optimistically nil-free carriers" group) pass unchanged.

## Performance

`rigor check lib --no-cache`. The arms alternate rep by rep, and the two changed files are toggled to `19d7105d` in-tree, with each file's hash printed per arm. Before (`d2f218af3ffb` / `c47244f5eb77`): 14.67, 14.20, 14.80, 14.60 s. After (`2b68d70c6a6d` / `528242fb228d`): 14.05, 14.44, 14.97, 15.17 s. Medians are 14.64 vs 14.71 s (+0.5%, within the spread). Both arms report the same diagnostics.

## Corpus gate

21 targets, `--no-cache`, cwd = target. Before = the changed files toggled to `19d7105d` inside this worktree. Every run parsed as non-empty JSON. Diagnostics before → after: liquid 5→5, kramdown 41→41, faraday 1→1, mail **23→24**, net-ssh 16→16, rubocop-ast 8→8, parser 5→5, concurrent-ruby 24→24, haml 21→21, slim 6→6, rgl 22→22, oj 1→1, ox 2→2, erubi 3→3, herb 511→511, jbuilder 5→5, protobuf 3→3, tdiary-core 247→247, redmine (app lib) 239→239, mastodon (app/models lib) 8→8, gitlab (app/models) 210→210. With `--no-baseline`: redmine 1016→1016, mastodon 316→316, gitlab 244→244. Teeth check: the same harness reports `assert.type-mismatch` for `assert_type("false | finite-float", x.finite? && x)` in the before arm only.

**The one moved diagnostic, adjudicated: a false positive from a pre-existing root cause, not fixed here.** mail `lib/mail/message.rb:1553` `if delivery_status_report?` now reports "always falsey". The root is overload selection. `Regexp#match?` is `(interned) -> bool | (nil) -> false`, and an argument typed `Dynamic[top] | nil` (`main_type`) is not treated as untyped by `OverloadSelector` (#521 guards only a bare `Dynamic[Top]`). The strict pass therefore pins `(nil) -> false`, and `multipart?` folds to `false`. The same root already fires this rule on master at `message.rb:1839` and `:1886` (`if multipart?`). On master, `rigor type-of` already reported `multipart_report? && …` (1541) as `false` in value position. This PR makes the `def` body's statement-position return agree with it, so the fold reaches a third site. A 7-line prototype that treats a union with an untyped member like an untyped argument in the strict, alias and #521 passes types `/re/.match?(Dynamic[top]?)` as `Dynamic[bool]`, and `spec/rigor/inference/method_dispatcher` stays green (1399 examples). It is left out of this PR because it changes overload selection for every nilable-untyped argument and needs its own corpus gate.

`make lint`, `make check`, `make check-plugins` (0 warnings/errors), `make docs-check`, `spec/rigor/inference` + `runner_check_rules_spec.rb` (3653), `spec/integration` (2175), and the remaining `spec/rigor/*` + `spec/docs` (6821; its only two failures were the provenance ratchet lowered above, re-run green) are green locally. `make verify` was not run in this lane, to avoid a concurrent full gate.


zonuexe:
Orchestrator decision: **held — not landing with the new `mail` false positive.**

The lane's analysis checks out on master `5922448f`: `mail/lib/mail/message.rb:1839` and `:1886` already report "condition is always falsey" on `if multipart?`, because `OverloadSelector` treats `Dynamic[top] | nil` as precise and picks `Regexp#match?(nil) -> false`. This PR makes `:1553` (`if delivery_status_report?`, body `multipart? && …`) agree with that wrong type, so it adds a third site of an existing false positive.

The fix for the cause removes false positives rather than moving them, so it lands first as its own change with its own corpus gate: #1021. This PR is then rebased onto it and re-gated; it lands once `mail` shows no diagnostic this PR adds. Staying Draft.

zonuexe:
Hold lifted. #1022 fixed the overload-selection cause (#1021) and is on master `5b3a8196`. With this branch merged onto it, a cold before/after against master shows no diagnostic added or removed on `mail` (19 → 19), `tdiary-core`, or `redmine` (236 → 236) — the `message.rb:1553` false positive this PR used to add no longer exists. `make verify` exit 0 and CI green on `b299e645`.

PR #1022, opened 2026-09-14
Treat a union with an untyped member as imprecise in overload selection

### LINKED ISSUE #1021: Overload selection treats Dynamic[top] | nil as precise, so Regexp#match?(maybe_untyped) types as false
## What happens

`OverloadSelector` treats an argument as untyped only when it is a bare `Dynamic[Top]`. An argument typed `Dynamic[top] | nil` — the ordinary shape of "an unknown value that may be nil" — is treated as precise, so the strict pass can pick an overload keyed on its `nil` arm alone.

`Regexp#match?` is declared `(interned) -> bool | (nil) -> false`. With `main_type : Dynamic[top] | nil`, the strict pass selects `(nil) -> false`, and the whole call types as the literal `false`. Everything downstream then believes a live condition is dead.

## Where it bites (survey corpus, master `5922448f`)

`mail` — `lib/mail/message.rb`:

```ruby
def multipart?
  has_content_type? ? /^multipart$/i.match?(main_type) : false
end
```

`rigor check --no-cache lib` on master reports `flow.always-truthy-condition` "condition is always falsey" at:

- `message.rb:1839` — `if multipart?`
- `message.rb:1886` — `if self.multipart?`

Both are false positives: `multipart?` is true for every multipart message the gem builds. PR #1020 (and/or value-position narrowing, #1016) makes a third site agree with the same wrong type — `message.rb:1553`, `if delivery_status_report?`, whose body is `multipart? && …` — so #1020 is held until this lands.

`message.rb:2059` (`if text? && !has_charset?`) may share the cause; `message.rb:2015` (`if @separate_parts`) almost certainly does not. Establish which, rather than assuming.

## A prototype exists

The #1016 lane wrote a 7-line prototype that treats a union with an untyped member as imprecise in all three places the selector checks `untyped_arg?` (the gradual fallback's all-matches return, the strict pass's `NO_MATCH`, and pass 1.5's core-alias guard). With it, `/re/.match?(Dynamic[top]?)` types as `Dynamic[bool]` and `spec/rigor/inference/method_dispatcher` stays green (1,399 examples). It changes overload selection for every nilable-untyped argument, so it needs its own corpus gate — that gate is this issue.

```ruby
def imprecise_arg?(type)
  untyped_arg?(type) || (type.is_a?(Type::Union) && type.members.any? { |member| untyped_arg?(member) })
end
```

Consider whether "a union with any untyped member" is the right boundary, or whether only `Dynamic | nil` (the optional-untyped shape) should qualify — a union like `Dynamic[top] | Integer` is also imprecise, but check what the gradual pass then returns for it.

## Acceptance criteria

- [ ] `/re/.match?(x)` with `x : Dynamic[top] | nil` no longer types as `false`; pinned by a spec that fails on master
- [ ] `mail`'s `message.rb:1839` and `:1886` false positives are gone, cold (`--no-cache`), and nothing new appears in `mail`
- [ ] A survey-corpus before/after over the same target set #1015 and #1020 used, every moved diagnostic adjudicated — this change is expected to REMOVE false positives, and any it adds must be explained
- [ ] Must-still-select specs: a precise `String` argument still picks `(interned) -> bool`, a literal `nil` still picks `(nil) -> false`, and #521's all-matches behaviour for a bare `Dynamic[Top]` is unchanged
- [ ] `make verify` green

## Blocked by

None - can start immediately. Blocks #1020 landing.

zonuexe:
Closes #1021

## Mechanism

`OverloadSelector` declined to discriminate only on a bare `Dynamic[top]` argument. A union with an untyped member (`Dynamic[top] | nil`, the ordinary "unknown value that may be nil") counted as precise, so the strict pass matched on the `nil` member alone: `Regexp#match?(maybe_untyped)` picked `(nil) -> false` and the call typed as the literal `false`.

## Boundary

- **Imprecise** = the bare `Dynamic[top]`, or a `Union` with a `Dynamic[top]` member. The untyped member may take any overload at runtime, so the strict pass and the alias-strict pass (1.5) skip, as they already did for a bare untyped argument, and the gradual pass returns every match for the #521 `Dynamic`-wrapped join.
- `Dynamic[top] | Integer` and `Dynamic[top] | String | nil` are included. Gradual acceptance still reads the whole union, so precise members keep excluding overloads they cannot reach: `Array#*` with `Dynamic[top] | Integer` now answers both `(string)` and `(int)` (it used to pin `(int)` by the alias pass, which the untyped member does not justify), `Integer#+` with it still answers only `(Integer) -> Integer`, and `nil`-pinned overloads drop out for `Dynamic[top] | Integer`. Narrowing the boundary to `Dynamic | nil` alone would leave the same position-decided pin for every other untyped union, with no corpus evidence that the wider boundary costs anything (zero added diagnostics below).
- The value-pinning exclusion in the gradual pass stays keyed on the bare `Dynamic[top]`. `nil` against `Dynamic[top] | nil` is accepted on real evidence (the nil member), so `(nil) -> false` stays a candidate and joins as `Dynamic[bool]`.
- A `Dynamic` with a non-`top` static facet (`Dynamic[String]`) is unchanged, as its facet discriminates. Note: probing showed `Regexp#match?(Dynamic[String])` also selects `(nil) -> false` on master and after this change, because `Constant[nil]` gradually "maybe"-accepts any `Dynamic`. That is an acceptance question, not this selector boundary, and is left for a separate issue.

Normative text: `docs/internal-spec/inference-engine.md` (the three-pass rule now names "imprecise" actuals, the alias-pass skip, and the #521 all-matches return).

## Specs

`overload_selector_spec.rb` gains 8 examples under "union with an untyped member". On master, 3 fail:

- `Regexp#match?` with `Dynamic[top] | nil`: expected `["::interned", "nil"]`, got `["nil"]`.
- `/re/.match?(y)` with `y = c ? x : nil` (asserted to type `Dynamic[top]?`): expected not `false`, got `Constant false`. It now types `Dynamic[bool]`.
- `Array#*` with `Dynamic[top] | Integer`: expected `["::string", "::int"]`, got `["::int"]`.

The 5 must-still-select controls pass on both engines: a precise `String` gives `(interned) -> bool`, a literal `nil` gives `(nil) -> false`, a bare `Dynamic[top]` keeps the #521 answer (`Regexp#match?` gives `(interned)`, `Array#*` gives both), a precise `String | nil` gives exactly one candidate (`Regexp#match?` and `Kernel#Array`), and `Dynamic[String]` gives exactly one candidate. `spec/rigor/inference/method_dispatcher` has 1,407 examples, 0 failures. The `sig/` provenance ratchet did not move.

## The mail sites

`rigor check --no-cache lib` in `mail`:

- `message.rb:1839` `if multipart?`: **gone**. `multipart?`'s `/^multipart$/i.match?(main_type)`, where `main_type` is `Dynamic[top] | nil`, types `false` on master and `Dynamic[bool]` after.
- `message.rb:1886` `if self.multipart?`: **gone**, same cause.
- `message.rb:2059` `if text? && !has_charset?`: **gone, same cause**. `text?` is `has_content_type? ? /^text$/i.match?(main_type) : false`. On master `text?` types `false`, and after the change `Dynamic[bool]`.
- `message.rb:2015` `if @separate_parts`: **gone, same cause, one hop further**. The ivar's writes are `@separate_parts = false` (line 110) and `@separate_parts = multipart?` (line 2123). On master the `multipart?` call at 2123 types `false`, so every write is `false`. After the change it types `Dynamic[bool] | false`.

## Corpus gate

21 targets, sequential, `--no-cache`, cwd = target, via the worktree's `exe/rigor` and Gemfile. "Before" = `overload_selector.rb` copied from `origin/master` inside this worktree (sha256 `b29bdd99249b`), "after" = the fix (`418251c5ae4b`), with a `trap` restoring the file. Every run exited normally with parseable non-empty JSON. After the gate, the committed file (`0746b97ae613`) differs only by an equivalent refactor (endless `untyped_arg?`, `||` in place of an early return, to stay inside `Metrics/ModuleLength`) plus comment reflow. The mail re-run on it shows the same four sites gone.

| target | before | after |
| --- | --- | --- |
| liquid | 5 | 5 |
| kramdown | 41 | 41 |
| faraday | 1 | 1 |
| mail | 23 | **19** |
| net-ssh | 16 | 16 |
| rubocop-ast | 8 | 8 |
| parser | 5 | 5 |
| concurrent-ruby | 24 | 24 |
| haml | 21 | 21 |
| slim | 6 | 6 |
| rgl | 22 | 22 |
| oj | 1 | 1 |
| ox | 2 | 2 |
| erubi | 3 | 3 |
| herb | 511 | 511 |
| jbuilder | 5 | 5 |
| protobuf (`ruby/lib`) | 3 | 3 |
| tdiary-core | 247 | **246** |
| redmine (`app lib`) | 239 | **236** |
| redmine `--no-baseline` | 1016 | **1013** |
| mastodon (`app/models lib`) | 8 | 8 |
| mastodon `--no-baseline` | 316 | 316 |
| gitlab (`app/models`) | 210 | 210 |
| gitlab `--no-baseline` | 244 | 244 |

Diagnostic records were compared field by field. No record was added or changed. 8 were removed, all `flow.always-truthy-condition` "condition is always falsey", and all are false positives:

- mail `message.rb:1839`, `:1886`, `:2015`, `:2059`: see above.
- redmine `lib/redmine/export/pdf.rb:152`, `if /^[^\/"]+\.(gif|jpg|jpe|jpeg|png|webp)$/i.match?(filename_utf8)`: `filename_utf8` comes from `CodesetUtil.to_utf8`, which returns nil or a String (`Dynamic[top] | nil`). Image attachments do match at runtime.
- redmine `pdf.rb:155`, `if atta && …`, and `pdf.rb:105`, `if atta` (in `get_image_filename`, fed by `RDMPdfEncoding.attach`): `atta` was assigned only inside the falsely dead branch at 152, so it folded to `nil`. Both are live.
- tdiary-core `lib/tdiary/style/tdiary.rb:71`, `s += '<' if /^</=~@subtitle`: the same shape on `Regexp#=~`. `@subtitle` is a String or nil, and the strict pass picked `(nil) -> nil`.

Teeth check, same toggle harness, on a scratch fixture (`if /^multipart$/i.match?(c ? x : nil)` through a helper method): the before arm reports `flow.always-truthy-condition`, and `type-of` gives `false`. The after arm reports no diagnostics, and `type-of` gives `Dynamic[bool]`.

## Performance

`rigor check lib --no-cache` on this repository. Arms alternate rep by rep, with the file toggled in-tree and hash-stamped:

| rep | before `b29bdd99249b` | after `418251c5ae4b` |
| --- | --- | --- |
| 1 | 14.37 s | 14.23 s |
| 2 | 15.39 s | 14.30 s |
| 3 | 14.28 s | 13.78 s |
| 4 | 14.64 s | 13.48 s |

The medians are 14.51 s before and 14.01 s after, which is within noise (the change adds one `is_a?` per argument per selection). The self-check output is identical across the two arms.

## Gates

`make lint`, `make check` (exit 0, no warnings), `make check-plugins` (exit 0), `make docs-check` (exit 0). `make verify` result is in the follow-up comment.


zonuexe:
`make verify` (run once, after the last edit including the changelog fragment): exit 0. `git diff --check`: clean.

PR #1023, opened 2026-09-14
Report the two inline-RBS spellings the reader still drops silently

### LINKED ISSUE #1019: Two inline-RBS spellings are still dropped silently: an @rbs-ext line and an Integer[1..10] parameter annotation
## What happens

Two inline-RBS spellings are still dropped by Rigor's reader with no diagnostic at all, after #997 (PR #1005) and #998 (PR #1018) closed the other silent rows of the probe table in `docs/notes/20260912-inline-refinement-carrier-probe.md`:

| row | spelling | what the `rbs-inline` gem does | Rigor today |
| --- | --- | --- | --- |
| B | `# @rbs-ext return: non-empty-string` above `# @rbs return: String` | `@rbs\b` matches before the hyphen, so it is an `@rbs` annotation with an unknown body; the gem swallows it | nothing reported; the `# @rbs return:` line beside it still binds |
| H | `# @rbs n: Integer[1..10]` | the parameter annotation yields a `VarType` with a nil type | `(untyped n)`, nothing reported |

rbs's built-in `RBS::InlineParser` reports both (`unexpected token for @rbs annotation`, `comma delimited type list is expected`), and so does Steep 2.0.0 in inline mode. Rigor is the only reader of the three that stays silent.

ADR-32 WD12 says a parsed-but-unhonoured annotation is never swallowed. Row H is the sharper one: the author wrote a parameter contract — in the refinement notation ADR-109 made canonical — and the method silently types as if they had written nothing.

## What to build

Route both through the existing `source-rbs-annotation-not-honoured` notice that #1005 and #1018 already use for the other rows, naming the line and the text that was not honoured.

- For **H**, the notice should say where `Integer[1..10]` IS valid (a `%a{rigor:v1:param: n is Integer[1..10]}` annotation beside a plain `Integer`), the way #1005's message does for a refinement name in a type position.
- For **B**, the notice should not suggest `@rbs-ext` is a recognised tag. ADR-111 (Proposed) measured it out as a name; if the maintainer later adopts a Rigor-owned tag, it will be spelled differently (`@extrbs` was measured clean in all three readers).

Pair each new notice with a must-still-succeed case: a well-formed `# @rbs n: Integer` and a well-formed own-line and same-line `%a{}` annotation stay notice-free.

## Acceptance criteria

- [ ] Rows B and H each produce a `source-rbs-annotation-not-honoured` notice naming the line and the dropped text, verified cold (`--no-cache`)
- [ ] Every other row of the probe table is unchanged — re-run the note's method and update its Rigor column
- [ ] Must-still-succeed specs for well-formed spellings
- [ ] `docs/manual/plugins/rigor-rbs-inline.md` lists the new causes
- [ ] `make verify` green

## Blocked by

None - can start immediately. Split out of #998: PR #1018 made the same-line `%a{}` forms bind, and these two rows are the remainder of #998's "no row dropped without a diagnostic" criterion.

zonuexe:
## Summary

Closes #1019, the remainder of #998's "no row is dropped without a diagnostic" criterion: two inline-RBS spellings still parsed and contributed nothing, with no diagnostic anywhere.

Row H — `# @rbs n: Integer[1..10]`: the gem's `# @rbs name: T` grammar makes the colon-and-type optional, so a `T` that does not parse leaves `VarType#type` nil rather than raising, and `Writer#default_type` fills the empty slot the same way an unannotated parameter is filled. The parameter reads `untyped n`, indistinguishable from a parameter nobody annotated.

Row B — `# @rbs-ext return: non-empty-string`: the gem's own `annotation_comment?` detector matches the word boundary right after `@rbs`, so this line is inside its net — the gem tries to read it as `@rbs` — but `parse_annotation`'s case has no branch for `-ext` and returns nil, folding the whole paragraph back into an ordinary `CommentLines`, the same class a comment that never claimed to be `@rbs` parses to.

Both now route through the existing ADR-32 WD12 `source-rbs-annotation-not-honoured` notice:

- H is detected off `RBS::Inline::AST::Annotations::VarType#type` being nil (the gem's own parse result), not a pattern over the comment. The notice names the line, the dropped text, and points at the `%a{rigor:v1:param: n is Integer[1..10]}` spelling that carries the refinement correctly.
- B is detected by matching the paragraph's `CommentLines` against `SameLineAnnotations::RBS_MARKER` — the exact `@rbs` marker the gem's own `annotation_comment?` already keys on, reused rather than duplicated — so a paragraph is flagged only when the gem's own detector would have called it an `@rbs` attempt. The message says only that the tag did not parse; it never implies `@rbs-ext` is a recognised tag.

## Before / after (measured cold, `--no-cache`, at `b73747d6`)

Row B (`# @rbs-ext return: non-empty-string` above `# @rbs return: String`):

Before: no diagnostic; the `# @rbs return: String` line still binds.

After: `plugin \`rbs-inline\` parsed an inline-RBS annotation in this file but did not honour it: a comment on line 3 (\`#@rbs-ext return: non-empty-string\`) opens with the \`@rbs\` marker but its body did not parse as a recognised \`@rbs\` annotation, so the whole line was DROPPED — this is not a recognised tag, and nothing on it was applied. Use one of the documented \`# @rbs\` forms (\`name: T\`, \`return: T\`, \`module-self T\`, …), or rename the tag so it does not start with \`@rbs\`, to avoid this notice. The file's other annotations are unaffected. [source-rbs-annotation-not-honoured]`

Row H (`# @rbs n: Integer[1..10]`):

Before: `(untyped n)`, no diagnostic.

After: `plugin \`rbs-inline\` parsed an inline-RBS annotation in this file but did not honour it: a \`# @rbs n:\` annotation on line 3 (\`Integer[1..10]\`) did not parse as an RBS type and was DROPPED — the parameter types as \`untyped\`, not merely as if its type were wrong. If \`Integer[1..10]\` is a Rigor refinement, write it as an \`%a{rigor:v1:param: n is Integer[1..10]}\` annotation beside the plain RBS type instead — see docs/manual/16-rbs-extended-annotations.md — to restore type coverage. The file's other annotations are unaffected. [source-rbs-annotation-not-honoured]`

Every other row of `docs/notes/20260912-inline-refinement-carrier-probe.md` was re-measured through the same Rigor-layer probe and is unchanged.

## Test plan

- [x] `make lint` — clean
- [x] `bundle exec rspec spec/integration/plugins/rbs_inline_plugin_spec.rb spec/rigor/effects/envelope_rbs_inline_spec.rb spec/rigor/environment/rbs_loader_spec.rb spec/rigor/analysis/runner/diagnostic_aggregator_spec.rb spec/integration/definition_build_failed_spec.rb spec/rigor/plugin/source_rbs_synthesis_reporter_spec.rb spec/rigor/bleeding_edge_spec.rb` — 315 examples, 0 failures
- [x] The four new row-B/row-H examples fail on the pre-fix `rbs_inline.rb` (verified by temporarily reverting just that file) and pass with the fix
- [x] `make check` / `make check-plugins` — clean
- [x] `make docs-check` — clean
- [x] Manually reproduced both fixtures cold (`rigor check --no-cache`) and confirmed the notice text above
- [x] `make verify` not run in this lane (another lane was running it concurrently; ADR-32/WD12-adjacent specs run individually above)


zonuexe:
Orchestrator review, cold (`--no-cache`) in a scratch project:

- **No false notice on valid rbs-inline syntax.** A file carrying `# @rbs use`, `generic`, `inherits`, a class-level `# @rbs @ivar:`, an `# @rbs!` block, `*args` / `**opts` / `&block` parameter annotations, `skip`, `override`, own-line `%a{pure}`, a `#:` method type and `module-self` produces no notice at all; only the deliberate `# @rbs n: Integer[1..10]` does.
- **Row B fires; its neighbours do not.** `# @rbs-ext return: …` gets the notice. A comment that mentions `@rbs` mid-sentence, a prose comment that begins with `@rbs`, and `# @extrbs …` all stay silent.

One cosmetic nit, not blocking: the row B notice quotes the line as `#@rbs-ext return: non-empty-string` — the space after `#` from the source (`# @rbs-ext`) is lost in the reconstructed text. Worth quoting the original comment slice if it is cheap.

PR #1137, opened 2026-09-20
Land references/ copies at the path instead of inside it

zonuexe:
`bin/rigor-worktree --with-references` nested every `references/` checkout one level deep — `references/ruby` became `references/ruby/ruby` — and the script's own verification could not detect it. A worktree made with the flag therefore ran `spec/docs/c_effects_raises_gate_spec.rb` as a **skip** ("references/ruby is an optional submodule (AGENTS.md) and is not checked out here") while `make docs-check` reported green: the silent SKIP the flag exists to prevent.

## Cause

`git worktree add` materializes an empty placeholder directory for every registered submodule path, and `cp -R SRC DST` with DST an existing directory copies **into** it. The placeholder is now removed before the copy, through a helper that can only ever remove a path inside the new worktree.

## Why nothing caught it

The check was `git rev-parse --verify HEAD` against the copy. The `.git` gitdir pointer is written at the correct level and resolves against the shared module store regardless of where the contents landed, so it passed on the nested copy and the WARNING branch never fired. `ls references/` and `git -C references/ruby rev-parse HEAD` both look correct on a broken copy, so neither is a usable verification either.

The check now compares the copy's top-level entries against the main clone's checkout and the script exits non-zero when any reference fails. Comparing contents rather than probing a path out of `git ls-tree HEAD` is what holds for the sparse checkouts: `references/phpstan` and `references/TypeScript-Website` deliberately materialize only part of HEAD, and a `ls-tree`-derived marker reported both *correct* copies as broken.

`clone_into_wt` gets the same guarded removal. `vendor/bundle` and `tool/*/vendor` are gitignored rather than registered submodules, so no placeholder exists for them today and they were never affected.

## Regression gate

`spec/tool/rigor_worktree_reference_clone_spec.rb` sources `reference_entries` and `verify_reference_clone` out of the script as shipped, so a rewrite that weakens the check is caught rather than mirrored. A verification that cannot fail is the defect, so the rejection cases are what get asserted — nested copy, partial copy — alongside the sparse-checkout acceptance. Both rejection examples fail against the old rev-parse-only check.

## Validation

- Throwaway worktree created with the fixed script and `--with-references`: all 13 checkouts land at their own paths, `references/ruby/object.c` present, `make docs-check` → **473 examples, 0 failures** (no pending).
- The same tree with `references/ruby` re-nested by hand → **473 examples, 0 failures, 1 pending**, gate skipped, while `ls` and `rev-parse HEAD` still report it correct and the new verification rejects it.
- All 13 references checked in both shapes: correct copy accepted, nested copy rejected, every one.
- `make verify` (12035 examples, 0 failures; lint 1331 files clean), `make docs-check`, `git diff --check`.

The `rigor-worktree` skill's `--with-references` bullet records that the script now verifies and exits non-zero, and why the cheap looks do not discriminate.


zonuexe:
## Adversarial review round

An adversarial review of the first commit found three defects. All three were reproduced before fixing; `6b3e86e6` fixes them.

**1. The `..` guard broke every run, not just `--with-references`.** `rm_wt_path` rejected `..` as a *substring*, and `wt_root` is never normalized. So `RIGOR_WT_ROOT=$PWD/../rigor-wt` — an ordinary spelling — died at the first `clone_into_wt vendor/bundle`, *after* `git worktree add` had registered the worktree, leaving a half-built tree that the re-run then refuses with `already exists`. This was a regression I introduced: on `master` the same invocation succeeds. Fixed by resolving `$wt` with `pwd -P` once the worktree exists, and rejecting `..` as a path *component* (which also stops an ordinary entry named `..foo` from tripping it).

**2. The script has never run inside the Nix Flake on macOS.** `cow_cp` picked `cp -Rc` on `uname = Darwin`, but the Flake AGENTS.md mandates puts GNU coreutils ahead of `/bin`:

```
nix develop --command bash -c 'command -v cp'   →  /nix/store/…-coreutils-9.11/bin/cp
nix develop --command bash -c 'cp -Rc a b'      →  cp: invalid option -- 'c'
```

Under `set -e` that aborts at the first copy. Pre-existing rather than introduced here, but it is why the first commit's end-to-end validation had to run on the host shell instead of through the Flake as the contract requires — so it is fixed here: probe `cp --version` first, fall back to `uname`.

**3. The new spec was not hermetic against the developer's git config.** It is the first spec in the suite to run `git commit`; a global `commit.gpgsign` or `core.hooksPath` failed the fixture and turned `make verify` red on that machine while the script under test was fine. Confirmed with `GIT_CONFIG_GLOBAL` carrying each — 4 examples, 4 errors. `Bundler.with_unbundled_env` covers Bundler's variables, not git's. Now pins `GIT_CONFIG_GLOBAL`/`GIT_CONFIG_SYSTEM`/`GIT_CONFIG_NOSYSTEM` for the fixture's git calls only, leaving the verification's own `git rev-parse` on the real config so the spec keeps testing what ships.

Also: SKILL.md claimed the script "checks each copy's contents"; it compares top-level entries, which is what it now says.

### Re-validation

Inside the Flake, with `RIGOR_WT_ROOT` carrying `..`, and `--with-references --with-tools` — the combination that exercises all three fixes at once:

```
RIGOR_WT_ROOT=…/rigor-wt/../rigor-wt nix develop --command ./bin/rigor-worktree \
  --with-references --with-tools <branch> origin/master
```

All 13 references and both tool bundles land; `references/ruby/object.c` present; `make docs-check` in the result → **473 examples, 0 failures**, no pending. The same invocation with `master`'s script dies at `cp: invalid option -- 'c'` with the worktree already registered.

Fail path re-proved after the changes: both rejection examples still fail against the old rev-parse-only check (4 examples, 2 failures), and pass against the current one.

`make verify` (12035 examples, 0 failures; lint clean over 1331 files), `make docs-check`, `git diff --check`.

### Reviewed and not broken

Glob metacharacters in `$wt` (the `case` pattern's `"$wt"` is literal); `$wt` or `$wt/` as the removal target (`?*` requires a character); a symlinked ancestor; `set -e` vs the spec's shell (every command in both functions sits in a conditional, a `||` list, a `case` or an assignment — identical verdicts with and without `-e`, so the spec tests what ships); `ref_failures=$((…))` is the assignment form, not the `((i++))` trap; the function-extraction regex fails loudly rather than silently on a rename or a truncated body; `spec/tool/` is picked up by `binpacker.yml`'s `test_pattern` and so runs in CI's sharded `make test-binpacker`.

Two residual limits, both judged acceptable and now stated rather than implied: the comparison is top-level-only, so a copy truncated deeper would pass (macOS `cp -Rc` exits non-zero on a partial copy, and `set -e` aborts before verification); and a top-level name beginning with `..` matches neither glob in `reference_entries` — no populated reference has one, and none has a symlink or a case-colliding name.

PR #1138, opened 2026-09-20
Pin the two docs gates that could not go red

zonuexe:
Two gates in `spec/docs/` generate their examples at load time from a glob, so a scan that matches nothing produces **no example** and the file reports green having checked nothing.

Measured on `master`:

| gate | corpus removed | result |
|---|---|---|
| `handbook_snippets_spec.rb` | `docs/handbook` | 20 examples → **0 examples, 0 failures** |
| `link_integrity_spec.rb` | `docs/` | 251 examples → **3 examples, 0 failures** |

Nothing downstream notices, because nothing asserts the count. CI's `Shard coverage (every test ran)` job checks that the shards agree on one partition of one suite — same shard count, same discovered count *across shards*, every index present — not that the suite is the size it was. All shards lose the same examples together, so they still agree.

## The reachable defect is the handbook gate

`extract_snippets` matches ```ruby and nothing else. A snippet written with any other Ruby-ish fence contributes no example and the count silently drops by one:

```
master, one counted snippet respelled to ```rb  →  19 examples, 0 failures
```

This is a live spelling hazard rather than a hypothetical one — `docs/handbook` already carries both ```rb and bare fences for non-executable blocks. Today all 20 executable blocks happen to be spelled ```ruby, so the new guard starts green; it exists for the next author who reaches for ```rb while writing an `assert_type` block.

The guard is deliberately narrow: only fences whose language token is absent, `rb`, or `ruby`. A ```text or ```sh block quoting CLI output may legitimately contain `assert_type(`, and failing a correct document is the error direction that matters (AGENTS.md: false positives outrank worst-case static reading).

## Link integrity is the lower risk of the two

Its corpus is `docs/**/*.md`, which cannot silently miss a new document. An attempt to reach its vacuity by breaking its own link extractor — making the inline-code stripper multi-line, a plausible "improvement" — made it **fail** rather than go quiet. It gets the size floor for symmetry: the same file already guards its self-referential scan this way (`it "scans a non-empty file set"`), and `packaged_link_integrity_spec.rb` guards both of its own corpora.

## Floors carry slack, on purpose

Following `plugin_io_boundary_spec.rb` (floor 50, actual 169) and `packaged_link_integrity_spec.rb` (70/200). A floor catches a collapse; the fence guard is what catches the loss of one snippet. Pinning today's exact count would go red on every legitimate edit, and the natural repair — lowering the number — silently reopens the hole.

## Each guard was verified to fail

| mutation | master | this PR |
|---|---|---|
| one counted snippet respelled ```rb | 20 examples, 0 failures | 21 examples, **1 failure** |
| `docs/handbook` hidden | 0 examples, 0 failures | 2 examples, **1 failure** |
| `docs/` hidden | 3 examples, 0 failures | 4 examples, **1 failure** |
| unmutated | — | 274 examples, 0 failures |

## Restructure, and why no example moved

The handbook file's helpers move to a `HandbookSnippets` module and its corpus becomes load-time constants, because the guards must read the scan from inside an example: `RSpec/LeakyLocalVariable` rules out the closure-local form, and a `def self.` helper is not callable from an example.

A `--dry-run` diff of every generated example description, before and after, is empty except for the three new guards — no existing example was renamed or lost.

## Validation

`make verify` (12045 examples, 0 failures; lint clean over 1333 files), `make docs-check` (476 examples), `git diff --check`.

## Surfaces checked and found sound

`manual_drift` (exits 1 on a missing corpus), `plugin_manual_examples` (`be >= 18`), `plugin_io_boundary` (`be > 50`), `packaged_link_integrity`, `type_shaped_comments`, `agent_index`, `llms_txt_drift`, `changelog_fragments` (inline validator examples pin the grammar while the directory is empty), `spec_analyzer_guard` (fixture arms pin the fail path), `changelog_conformance`, `effect_diagnostic_taxonomy`. `bin/doctor-submodules` exits 1 when it finds a problem. No `WARNING`-without-exit remains in `bin/`; no `-` prefixed recipes or exit-losing pipes in the Makefile. `plugin_contract_conformance` loops 45 directories inside 3 examples, so it cannot silently drop a new plugin.


zonuexe:
## Adversarial review round — NO-GO on the previous approach, now fixed

An adversarial review refuted the central design claim of `44003a51`: a floor with slack plus a three-spelling fence guard does **not** close the vacuity. Everything below was reproduced against that commit before changing it.

### 1. The floor could not detect any loss short of the directory vanishing

`>= 15` against an actual of 20, where the largest handbook file holds **four** snippets:

| mutation against `44003a51` | result |
|---|---|
| delete `04-tuples-and-shapes.md` (4 snippets) | 18 examples, **0 failures** |
| delete five snippet-bearing files | 17 examples, **0 failures** |
| respell all four fences in one file to ```console | 18 examples, **0 failures** |

### 2. The slack was cargo-culted from a corpus that grows

`plugin_io_boundary_spec.rb` floors 169 files at 50 because every new plugin adds to it. The handbook snippet count is **20 at every one of the last fourteen release tags**, across 33 commits touching `docs/handbook`:

```
v0.2.6 … v0.3.9  →  20 20 20 20 20 20 20 20 20 20 20 20 20 20
```

So the PR body's "pinning today's exact count would go red on every legitimate edit" was wrong — it would have churned **zero** times. It is now `eq(20)`.

### 3. The fence guard had walk-through holes, and one stray fence disabled it

Each of these is a `dump_type(1)` block that `44003a51` neither ran nor flagged: ```irb, ```console, `~~~ruby`. Worse, the guard's scan anchored its closers at column 0 while the extractor's did not — **the two disagreed about what a fence is**, exactly where the guard was supposed to cover the extractor's blind spot. A single indented closer above a ```rb block left it unmatched and the guard passed.

Both uses now share **one line scanner**, and the filter is an allow-list (`text`, `sh`, `diff`) rather than a list of known mis-spellings. The scanner accepts `~~~` as a marker, so a `~~~ruby` snippet is run rather than flagged.

### 4. Link integrity's real knob is the exclude, not the glob

Adding `manual/` to `LINK_INTEGRITY_EXCLUDE` took the scan from 248 documents to 200 with **0 failures**. The floor is raised to 230 (a loose one detected only the loss of `docs/adr/`, 111 of 248) and a second example now states what the exclusion may cover.

### The two mechanisms are complementary, and both are needed

The exact pin catches a snippet that **stops** being counted. It cannot catch one that was **never** counted — a newly written block fenced ```irb leaves the total at 20 while going unverified. That is the fence guard's job, and why it had to actually work.

### Re-verified matrix

| mutation | `44003a51` | now |
|---|---|---|
| existing snippet → ```rb / ```console / ```irb | 19–18 examples, 0 failures | 21 examples, **2 failures** |
| delete a four-snippet handbook file | 18 examples, 0 failures | 20 examples, **1 failure** |
| hide `docs/handbook` | 0 examples, 0 failures | 2 examples, **1 failure** |
| NEW block ```irb / ```console | 22 examples, 0 failures | 22 examples, **1 failure** |
| NEW block `~~~ruby` | 22 examples, 0 failures | 23 examples, **1 failure** |
| indented closer, then ```rb | 22 examples, 0 failures | 22 examples, **1 failure** |
| `manual/` added to the exclude | 251 examples, 0 failures | 201 examples, **2 failures** |
| hide `docs/` | 3 examples, 0 failures | 5 examples, **1 failure** |
| ```text quoting a diagnostic that mentions `assert_type(` | — | 22 examples, 0 failures (no false positive) |
| unmutated | — | 275 examples, 0 failures |

### Corrections to the original PR body

- "Pinning today's exact count would go red on every legitimate edit" — **false**, see above.
- "`Shard coverage` compares the discovered count across shards" — true, but `binpacker`'s `discovered_tests` is a count of **files**, not examples, so no whole-suite example baseline was cheaply available. The conclusion (nothing asserts the example count) holds.
- `make docs-check` is 477 examples **with 1 pending** — the pending is the C-effects gate, which needs the optional `references/ruby` checkout.

### Validation

`make verify` (12046 examples, 0 failures; lint clean over 1333 files), `make docs-check` (477 examples), `git diff --check`. The dry-run description diff against master is still empty except for the four new guards.

### Raised and deliberately not changed

- The bare-fence arm of the guard is the one case that could fail a correct document (a bare block quoting a diagnostic that mentions `assert_type(`). Kept, because a bare block holding `assert_type(` is far more likely to be a snippet that lost its language tag, and the repair is one word. Documented in place.
- Floors on `c_effects_raises_gate_spec.rb` (live catalogue glob) and `require_hygiene_spec.rb` — real, same shape, out of scope here.
- The `**[dev]**` changelog label is thinly precedented (`#94`, and `#1137`). Kept as the accurate one for contributor-facing tooling.

zonuexe:
## Second adversarial review round — NO-GO again, now fixed

A second review pass, on the *new* design, found that round 2's thesis did not hold. **Five realistic editing accidents still produced green-having-checked-nothing with the count pinned at 20.** All were reproduced against `cb2288ac` before being fixed.

### The hole both new mechanisms missed: the gate never checked that a snippet PARSED

The per-snippet example filtered `assert.type-mismatch` and nothing else. A snippet that does not parse never evaluates its own `assert_type`, so the filter finds nothing and the example passes:

| mutation against `cb2288ac` | result |
|---|---|
| flip an assertion to a wrong type **and** append `def (` | 22 examples, **0 failures** |
| typo one closer as ```` ```ruby ```` (two blocks merge into a 144-line snippet, 3 assertions, 82 unparsed diagnostics) | 22 examples, **0 failures** — count stays exactly 20, because the swallowed block was not itself executable |

A parse failure surfaces as a diagnostic carrying no rule. That is now checked **first and separately**.

### `executable?` was a substring test over the whole block

Commenting out every assertion left the block counted, running and green: 22 examples, **0 failures**. It is now line-wise and comment-aware.

This is not a hypothetical edit. `6a5225a1` took the corpus from 24 to 20 by **deleting** `assert_type` lines so failing snippets would pass; commenting them out instead would have moved no number that this PR pins.

### The scanner dropped an unclosed fence at EOF

The silent direction — neither run nor audited. ```` ```ruby ```` and ```` ```irb ```` blocks left unterminated: 22 examples, **0 failures** each. The block is now kept, so it reaches the audit, the parse check, or the pin; whichever it reaches is loud.

### Link integrity

- A floor of 230 against 248 left exactly enough slack to drop `docs/type-specification/` (18, the **binding** spec) or `docs/design/` (17) and land on the boundary green. Raised to **240**.
- Counting documents cannot see through the extractor itself: relaxing the fenced stripper to `/```.*```/m` took links actually checked from **2899 to 2439** while only 7 documents dropped out — every floor still passed. A second floor now counts **links**, not documents.

### Re-verified

| mutation | `cb2288ac` | now |
|---|---|---|
| wrong assertion hidden by a syntax error | 22 / 0 | 22 / **1** |
| all assertions commented out | 22 / 0 | 20 / **1** |
| unclosed ```` ```ruby ```` at EOF | 22 / 0 | 23 / **1** |
| unclosed ```` ```irb ```` at EOF | 22 / 0 | 22 / **1** |
| closer typo'd to ```` ```ruby ```` | 22 / 0 | 22 / **1** |
| greedy fence stripper in the extractor | 246 / 0 | 247 / **1** |
| `type-specification/` excluded (the two-line edit) | 235 / 0 | 236 / **2** |
| existing snippet respelled ```` ```rb ```` | — | 21 / **2** |
| delete a four-snippet file | — | 18 / **1** |
| hide `docs/handbook` | — | 2 / **1** |
| ```` ```text ```` quoting a diagnostic mentioning `assert_type(` | — | 22 / **0** (no false positive) |
| unmutated | — | 22 / 0 and 254 / 0 |

### Corrections to my earlier claims in this PR

- **"A static corpus … an exact pin has never had anything to churn on" is false.** The count changed eight times between 2026-05-07 and 2026-06-11 — 17 → 19 → 21 → 22 → 17 → 18 → 20 → 24 → 20 — and has been 20 since. The fourteen-tag window I cited **starts sixteen days after the last change**. The exact pin is still the right call, because the number belongs in the diff, but the corpus is quiet rather than static.
- The round-2 table's "delete a four-snippet handbook file → 20 examples" was mislabelled: that run deleted a **two**-snippet file. Deleting the four-snippet file gives **18 examples, 1 failure**.
- "33 commits touching `docs/handbook`" is 31.

### Validation

`make verify` (12047 examples, 0 failures; lint clean over 1333 files), `make docs-check` (478 examples, 1 pending — the C-effects gate, which needs the optional `references/ruby` checkout), `git diff --check`.

### Raised and deliberately not changed

- **Snippet bodies are byte-identical** between master's regex and the line scanner — all 20 `(file, index, body)` triples were extracted under each implementation and SHA-compared. This closes the real gap in my earlier verification, which compared example *descriptions* only and so proved nothing about the source each example analyses.
- A four-backtick wrapper containing a ```` ```ruby ```` snippet is flagged with advice that would be wrong for a page *showing* how to write a snippet. No such page exists; noted rather than pre-solved.
- Blockquoted and BOM-prefixed fences are silent to the scanner, where master's regex at least attempted them. Neither occurs in the handbook.
- Floors on `c_effects_raises_gate_spec.rb` and `require_hygiene_spec.rb` remain out of scope for this PR.

zonuexe:
## Third adversarial review round — conditional GO, conditions met

The third pass found **two more silent paths, both in the lexical tests round 3 introduced**, and judged the severity trend down: 0 severe, 2 moderate, 5 marginal. No one-step edit on today's corpus lands green; both new paths need a precondition the handbook does not currently meet.

Its recommendation was to **replace the proxies rather than special-case them** — each lexical rule answers one spelling, a parser answers the class. Taken. The result is less code than it replaces.

### `d.rule.nil?` was a proxy for "did not parse", and the analyzer has a path that emits nothing

`Runner#analyze_file_body` consults `ErbTemplateDetector` when a parse fails, and that detector is a byte-level test for `%>`. So:

| mutation against `410275c3` | result |
|---|---|
| wrong assertion + `def (` | 22 / 1 |
| wrong assertion + `def (` + a string holding `"<%= x %>"` | 22 / **0** |

Round 2's hole, silent again, behind one Rails-flavoured string. `Prism.parse` is now asked directly: it bypasses the analyzer and cannot be silenced by it.

### `executable?` skipped `#` lines, so `=begin`/`=end` kept a block counted and green

22 examples, **0 failures** — the same editing motion round 3 fixed, spelled differently. A trailing comment, a heredoc, a string literal and `__END__` are the same class. It now walks the AST for a receiverless call to `assert_type`/`dump_type`, sharing its parse with the check above.

A block that does not parse still yields a partial tree, so a snippet broken by a typo stays counted and reaches the parse check rather than quietly leaving the corpus.

### Also

- The count pin's comment still carried the "static corpus … never had anything to churn on" rationale I retracted in this PR but not in the file. It now states the eight changes and their dates.
- Link floor 2800 → 2700: every extractor change that drops a *shape* of link is red at either number (tables −408, list items −614, `..` paths −1336, greedy stripper −449), while 2800 also tripped on a legitimate rewrite of `docs/adr/README.md`, which alone carries 116 links.
- The changelog fragment described round 1; it now describes what ships, including the parse check — a new failure mode for a handbook snippet.

### Full matrix, all four rounds

| mutation | before | now |
|---|---|---|
| wrong assertion + syntax error + a `%>` string | 22 / 0 | 22 / **1** |
| assertions wrapped in `=begin`/`=end` | 22 / 0 | 21 / **1** |
| wrong assertion hidden by a syntax error | 22 / 0 | 22 / **1** |
| all assertions commented out | 22 / 0 | 20 / **1** |
| unclosed ```` ```ruby ```` at EOF | 22 / 0 | 23 / **1** |
| unclosed ```` ```irb ```` at EOF | 22 / 0 | 22 / **1** |
| closer typo'd to ```` ```ruby ```` | 22 / 0 | 22 / **1** |
| existing snippet respelled ```` ```rb ```` | 19 / 0 | 21 / **2** |
| NEW block fenced ```` ```irb ```` | 22 / 0 | 22 / **1** |
| delete a four-snippet file | 18 / 0 | 18 / **2** |
| hide `docs/handbook` | 0 / 0 | 2 / **1** |
| greedy fence stripper in the link extractor | 246 / 0 | 247 / **1** |
| ```` ```text ```` quoting a diagnostic that calls `assert_type` | — | 22 / **0** (no false positive) |
| unmutated | — | 22 / 0 and 254 / 0 |

### Validation

`make verify` (12047 examples, 0 failures; lint clean over 1333 files), `make docs-check` (478 examples, 1 pending — the C-effects gate, which needs the optional `references/ruby` checkout), `git diff --check`.

### Convergence

Round 1 and round 2 each broke on **one-step** edits to existing content. Round 3's two paths needed a conjunction — an accident *plus* content the handbook does not have. The mechanisms themselves (exact pin, single scanner, retained EOF block, allow-list audit) held against every direct attack across three passes; what gave way each time was a **proxy**: "did it parse" approximated by a diagnostic filter, "does it assert" approximated by a text search. Both are now answered by the parser, which is why this round removed code rather than adding a case.

The reviewer's own estimate was that a fourth pass returns only cosmetic findings **if** the repair is the replacement rather than another special case, and would find the next spelling (`__END__`, line continuations) if it were another patch. The replacement is what landed.

